PR #80, opened 2021-05-19
Use pinned version of nfpm via go modules

### LINKED ISSUE #76: Use pinned version of nfpm
We currently use the [latest stable version](https://github.com/upbound/up/blob/453b18b20de72ae3a3194ca481f5a0e128ba0f18/Makefile#L42) of `nfpm` in CI. We should instead pin to a specific version and update when appropriate. It appears the install script doesn't currently allow selecting a version for install, but we can download directly from the GitHub release if needed.

hasheddan:
Adds a build package that pins nfpm and invokes it via go generate so
that we can track its version in our go.mod.

Fixes #76


internal/build/build.go
```diff
@@ -0,0 +1,27 @@
...
+package build
+
+// NOTE(hasheddan): See the below link for details on what is happening here.
+// https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
+
+//go:generate go run github.com/goreleaser/nfpm/v2/cmd/nfpm pkg --config $CACHE_DIR/nfpm.yaml --packager $PACKAGER --target $OUTPUT_DIR/$PACKAGER/$PLATFORM/up.$PACKAGER
+
+import (
+	_ "github.com/goreleaser/nfpm/v2" //nolint:typecheck
+	_ "github.com/spf13/cobra/doc"    //nolint:typecheck
```
hasheddan, line 26: The `cobra` dependency was required to get `go.mod` not to complain about a missing transitive dependency

muvaf (approved):

PR #85, opened 2021-05-24
Strip leading v off of UXP version when pulling chart

### LINKED ISSUE #84: `up uxp install / upgrade` fails if `v` is supplied in version
Running `up uxp install v1.2.1-up.3` / `up uxp upgrade v1.2.1-up.3` fails because helm strips off the `v` when storing the chart in the cache:
```
up: error: stat /home/dan/.cache/up/charts/universal-crossplane-v1.2.1-up.3.tgz: no such file or directory
```
Because UXP uses a leading `v` in its versioning scheme and it is valid [semver](https://semver.org/), we should support specifying the version to install with a leading `v`.

hasheddan:
Strips the leading v off of any UXP versions supplied when pulling the
chart so that it matches how Helm caches it. An alternative approach
would be to parse into structured semantic version prior to passing to
Helm, but we prefer to defer that logic to Helm rather than parsing into
semver twice. If additional modification were made by Helm in the
future, this functionality could break, but we would prefer to hit the
issue rather than potentially pulling and installing incorrect UXP versions.

Fixes #84

### DIFF
--- a/internal/uxp/installers/helm/helm.go
+++ b/internal/uxp/installers/helm/helm.go
@@ -19,6 +19,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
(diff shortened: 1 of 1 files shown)


AaronME (approved):

hasheddan (commented):
Tested this successfully with:

```
up uxp install v1.2.1-up.2
up uxp upgrade v1.2.1-up.3
```

PR #90, opened 2021-05-26
Remove kube cluster ID from attach and defer to agent

### LINKED ISSUE #89: Need a way to create self-hosted control planes without kube-uid

In some scenarios, kube-cluster uid is not known or easily accessible by the user but available to agent running on the cluster. 

Deployments through marketplaces are a good example here. For example, while deploying through Rancher marketplace, all deployment flow can be completed through UI without connecting cluster from a local terminal (with up installed).

### How could up cli help solve your problem?

It should be possible to create a self-hosted control plane and obtain a **control plane token** without cluster connectivity or providing kube-uid to the cli.

Or to unify the experience, we can always leave setting kube-uid to the agent (with the first connection attempt) and `up cloud controlplane attach $NAME` command never sets kube-uid.

hasheddan:
Removes setting the kube cluster ID for a control plane when attaching
self-hosted and instead defers the responsibility to the agent upon
connecting.

Fixes #89 

Tested with normal `up cloud xp attach <name> | up uxp connect -` flow and successfully provisioned and connected a self-hosted control plane.

### DIFF
--- a/cmd/up/cloud/controlplane/attach.go
+++ b/cmd/up/cloud/controlplane/attach.go
@@ -19,10 +19,7 @@ import (
 	"fmt"
 
 	"github.com/alecthomas/kong"
-	"github.com/google/uuid"
 	"github.com/pkg/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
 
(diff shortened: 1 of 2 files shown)


turkenh (approved):
Thanks @hasheddan for this very quick fix 🙌

PR #100, opened 2021-06-22
Cache latest chart with proper version on install and upgrade

hasheddan:

Currently, when a version is not specified on uxp install or upgrade, we
pull the latest stable chart and cache it without version information in
the dedicated uxp chart cache. This means that if a version that was
installed as the latest was subsequently installed explicitly (i.e. by
providing the version), we would unnecessarily pull the chart again when
it was already in the cache.

This updates the pull logic to save charts pulled as latest with their
appropriate version so that they can be identified in future installs.

I have:

I removed the chart cache (`rm -rf ~/.cache/up/charts`), then ran `up uxp install` (with no version specified) and observed the proper versioned chart in the cache:

```
🤖 (up) ls ~/.cache/up/charts/
universal-crossplane-1.2.2-up.1.tgz
```

### DIFF
--- a/internal/uxp/installers/helm/helm.go
+++ b/internal/uxp/installers/helm/helm.go
@@ -352,46 +352,54 @@ func (h *installer) Uninstall() error {
 }
 
 // pullAndLoad pulls and loads a chart or fetches it from the catch.
-func (h *installer) pullAndLoad(version string) (*chart.Chart, error) {
-	// helm strips versions with leading v, which can cause issues when fetching
-	// the chart from the cache.
-	version = strings.TrimPrefix(version, "v")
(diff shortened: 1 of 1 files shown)


internal/uxp/installers/helm/helm.go
```diff
@@ -352,46 +352,54 @@ func (h *installer) Uninstall() error {
 }
 
 // pullAndLoad pulls and loads a chart or fetches it from the catch.
```
AaronME, line 354:
```suggestion
// pullAndLoad pulls and loads a chart or fetches it from the cache.
```

AaronME (approved):
LGTM

PR #107, opened 2021-07-15
Support for installing/upgrading charts from local files & dirs

ulucinar:

This PR adds support for installing/upgrading crossplane (or uxp) charts from a local  chart archive or a local directory. The local chart to be processed can be specified with the `--bundle` command-line option.

Note: I'm planning to use the changes in this PR in the demo, and I believe it could prove to be useful for development scenarios. If this functionality is already provided by other means, or if you believe it's not appropriate, we can close this PR.

I have:

Manually tested installing a downloaded UXP chart both from an archive and from a folder containing extracted contents. Also  verified installations & upgrades from a Helm repo still work.

### DIFF
--- a/cmd/up/uxp/common.go
+++ b/cmd/up/uxp/common.go
@@ -1,0 +1,25 @@
+// Copyright 2021 Upbound Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package uxp
+
+import "os"
+
+// ChartParams common parameters for installing/upgrading charts
+type ChartParams struct {
+	Unstable bool              `help:"Allow installing unstable UXP versions."`
+	Set      map[string]string `help:"Set Helm parameters."`
+	File     *os.File          `short:"f" help:"Helm parameters file."`
+	Chart    *os.File          `help:"Chart archive file."`
+}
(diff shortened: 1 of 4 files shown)


cmd/up/uxp/common.go
```diff
@@ -0,0 +1,25 @@
...
+
+package uxp
+
+import "os"
+
+// ChartParams common parameters for installing/upgrading charts
+type ChartParams struct {
+	Unstable bool              `help:"Allow installing unstable UXP versions."`
+	Set      map[string]string `help:"Set Helm parameters."`
+	File     *os.File          `short:"f" help:"Helm parameters file."`
```
hasheddan, line 23: I think we kept `Helm` out of the descriptions for these because we anticipated we may want to use them for other installation methods that we support in the future. What would you think about:
```suggestion
	Set      map[string]string `help:"Set parameters."`
	File     *os.File          `short:"f" help:"Parameters file."`
```

hasheddan (commented):
@ulucinar thanks for working on this! We had previously been somewhat hesitant to add this type of functionality because of the potential to support other installer backends than Helm. However, I think this is a very useful feature and don't want to hold off on its implementation due to the potential of supporting something else in the future :) I left one comment below, but I was also wondering if you felt making `--chart` more generic, maybe something like `--bundle` or `--src` would be an improvement or just more confusing? If we kept it as `--chart` that is obviously specific to Helm, but the alternatives are more generic and could be supported for other installers. On the other hand, we could just say `--chart` is only valid if using the Helm installer. What do you think?

cmd/up/uxp/common.go
ulucinar: Done.

ulucinar:
Hi @hasheddan,
Thanks for mentioning about the intention to hide Helm beneath. I did some changes in the wording and renamed `--chart`, which imposes Helm as you mentioned, as `--bundle`. 

If, in the future, a new bundle format that we adopt does not somehow support installation/upgrades from a local path, then we may have to remove this flag but I think this is fair.

hasheddan (approved):
Thanks for adding this awesome functionality @ulucinar and my apologies for the delayed review / merge. Excited to get this released to folks :)

PR #120, opened 2021-10-01
Create image pull secret on enterprise install

### LINKED ISSUE #119: Create image pull secrets on enterprise install

Currently, when a user runs `up enterprise install` they are prompted for a `License ID` and `Token`. These are then used to fetch an access key from DMV and download the appropriate Helm chart from Upbound Registry. However, the images referenced in the Helm chart will also need to be able to be pulled after chart install. The same License ID and Token can be used as an `imagePullSecret` to pull the images.

`up enterprise install` should create the necessary `Secret` used to pull images referenced in the Helm chart. We should also provide a flag for users to skip creating this `Secret` as there is a good chance many enterprise customers will be copying the images to their own internal registries, in which case they will need to supply their own image pull `Secret` manually. We may support prompting for these credentials at install in the future via some of the ideas expressed in #113.

Epic #

hasheddan:

Updates the up enterprise install command to create an image pull secret
using the provided license id and token. The image pull secret is
required to fetch the component images included in the enterprise helm
chart.

Fixes #119 

I have:

I used the following command to install a development version of `enterprise` and verified that credentials were created correctly and images were pulled sucessfully.
```
up enterprise install <version> --repo <redacted> --skip-license --set mysql.mysqlUser=upbound --set mysql.mysqlPassword=upbound --set redis.auth.password=supersecret
```

### DIFF
--- a/cmd/up/enterprise/install.go
+++ b/cmd/up/enterprise/install.go
@@ -16,6 +16,8 @@ package enterprise
 
 import (
 	"context"
+	"encoding/base64"
+	"encoding/json"
 	"io"
 	"net/url"
 	"time"
@@ -25,6 +27,7 @@ import (
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/kubectl/pkg/cmd/create"
 	"sigs.k8s.io/yaml"
 
 	"github.com/upbound/up/internal/auth"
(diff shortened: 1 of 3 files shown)


cmd/up/enterprise/install.go
```diff
@@ -198,18 +212,49 @@ func (c *installCmd) applyLicense(ctx context.Context, ns string) error {
...
+		_, err = c.kClient.CoreV1().Secrets(ns).Update(ctx, secret, metav1.UpdateOptions{})
 	}
 
 	return err
 }
+
+func (c *installCmd) applyImagePullSecret(ctx context.Context, ns string) error {
+	regAuth := &create.DockerConfigJSON{
+		Auths: map[string]create.DockerConfigEntry{
+			c.Registry.String(): {
```
hasheddan, line 226: Debated whether we should make it such that the image pull secret registry could differ from the `enterprise` registry location, but we would then find ourselves in the situation where we would need to prompt for alternate credentials, so I don't think it is worth it. We could add a `--skip-pull-secret` flag to allow for _not_ creating the secret in the event that a customer had pulled component images manually and pushed them to an internal registry, but they can already create a separate image pull secret and reference it via the values they pass to the helm chart, so I opted to leave this as always getting created for now and we can revisit in the future if needed.

tnthornton (approved):

PR #134, opened 2022-01-22
Add showMessage support to `xpls` and new version notifications

tnthornton:
Add support to `xpls` for emitting messages to the end user through `window/showMessage`. For now, this is only used to inform the end user that a new version is available to download up.

I have:

Beyond the unit tests, I hardcoded my version locally to "v0.5.0" and confirmed that I received the following pop-up:
Screen Shot 2022-01-21 at 4 04 25 PM


internal/xpls/version/version.go
```diff
@@ -0,0 +1,140 @@
...
+		return "", "", false
+	}
+
+	return local, remote, i.newAvailable(local, remote)
+}
+
+func (i *Informer) newAvailable(local, remote string) bool {
+	lv, err := semver.NewVersion(local)
+	if err != nil {
+		// invalid remote version detected
```
hasheddan, line 97:
```suggestion
		// invalid local version detected
```
should this be in error message?

internal/xpls/version/version.go
```diff
@@ -0,0 +1,140 @@
...
+	"io/ioutil"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/Masterminds/semver/v3"
+
+	"github.com/crossplane/crossplane-runtime/pkg/logging"
+
+	"github.com/upbound/up/internal/version"
```
hasheddan, line 29: Could this be bundled into the existing `version` pkg? I could see this being useful for checking for a new available version in `up -v` :) (perhaps it should be an optional flag though so that it doesn't make an unexpected network call -- could explore in the future)

tnthornton: Sure thing 👍 (I'll move it). I'll hold off on adding it to `up --version` for now though.

hasheddan: SGTM 👍🏻

internal/version/version.go
```diff
@@ -14,9 +14,133 @@
...
 var version string
 
 // GetVersion returns the current build version.
 func GetVersion() string {
 	return version
 }
+
+type client interface {
+	Do(*http.Request) (*http.Response, error)
+}
```
hasheddan, line 49: Probably wouldn't hurt for us to have a small `http` package eventually as we are probably doing (or will do) some of this in a variety of places 👍🏻

hasheddan (approved):
LGTM

internal/version/version.go
tnthornton: Totally agree.

PR #145, opened 2022-01-31
Add support for validating XRD's openAPIV3Schema

tnthornton:
This change set adds the following new features to `xpls`:
- initial support for validating the openAPIV3Schema definition in an XRD using CRD's validation mechanisms
- adds functionality to warn the end user if a validator cannot be found for a GVK

In addition, the go mod minimum version requirement was bumped in order to account for repo name changes in github.com/googleapis/gnostic that transitive dependencies weren't able to handle very well.

Fixes #https://github.com/upbound/up-ls/issues/59

I have:

Beyond unit tests, general functional tests were performed to verify we saw the expected warns and errors. You can see in the screenshots below what that looks like:

### `openAPIV3Schema` validation:
Screen Shot 2022-01-31 at 12 28 19 PM

### warn end user if GVK is missing due to XRD's definition being broken:
Screen Shot 2022-01-31 at 1 47 09 PM


go.mod
```diff
@@ -1,53 +1,227 @@
...
+	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
 )
 
 replace (
 	github.com/crossplane/crossplane/controller/apiextensions => ./internal/vendor/github.com/crossplane/crossplane/controller/apiextensions
+	github.com/crossplane/crossplane/xcrd => ./internal/vendor/github.com/crossplane/crossplane/xcrd
 	github.com/docker/distribution => github.com/docker/distribution v0.0.0-20191216044856-a8371794149d
 	github.com/docker/docker => github.com/moby/moby v17.12.0-ce-rc1.0.20200618181300-9dc6525e6118+incompatible
 	github.com/golang/tools => ./internal/vendor/golang.org/x/tools
+	github.com/googleapis/gnostic => github.com/google/gnostic v0.5.5
```
tnthornton, line 226: See note about needing to go to `1.17`. This change was rooted in updating `k8s.io/apiextensions-apiserver` to `v0.23.3`.

internal/xpkg/snapshot/validators.go
```diff
@@ -157,6 +161,15 @@ func validatorsFromV1Beta1XRD(x *xpextv1beta1.CompositeResourceDefinition, acc m
 }
 
 func validatorsFromV1XRD(x *xpextv1.CompositeResourceDefinition, acc map[schema.GroupVersionKind]*validator.ObjectValidator) error {
+	errs := validateOpenAPIV3Schema(x)
+	if errs != nil {
```
hasheddan, line 165:
```suggestion
	if len(errs) != 0 {
```

Current implementation panics on zero-length slice.

internal/xpkg/snapshot/xrd.go
```diff
@@ -0,0 +1,185 @@
...
+
+// if the validations were all move to spec.validation, update the path
+// to point to spec.version[0]
+func replaceValidation(fieldVal string) string {
+	return strings.Replace(fieldVal, "spec.validation", "spec.versions[0].schema", 1)
+}
+
+// paths are return from CRD validations using map[key].field notation
+func replaceMapKeys(fieldVal string) string {
+	sampleRegexp := regexp.MustCompile(`(\[([a-zA-Z]*)\])`)
```
hasheddan, line 176: I believe we can avoid calling this on every invocation of `replaceMapKeys` (i.e. compile once then use it where necessary)

internal/xpkg/snapshot/xrd.go
```diff
@@ -0,0 +1,185 @@
...
+	for _, f := range fns {
+		cleaned = f(cleaned)
+	}
+
+	return cleaned
+}
+
+type cleaner func(string) string
+
+// if the validations were all move to spec.validation, update the path
```
hasheddan, line 168:
```suggestion
// if the validations were all moved to spec.validation, update the path
```

internal/xpkg/snapshot/xrd.go
```diff
@@ -0,0 +1,185 @@
...
+// to point to spec.version[0]
+func replaceValidation(fieldVal string) string {
+	return strings.Replace(fieldVal, "spec.validation", "spec.versions[0].schema", 1)
+}
+
+// paths are return from CRD validations using map[key].field notation
```
hasheddan, line 174:
```suggestion
// paths are returned from CRD validations using map[key].field notation
```

internal/xpkg/snapshot/xrd.go
```diff
@@ -0,0 +1,185 @@
...
+func replaceMapKeys(fieldVal string) string {
+	sampleRegexp := regexp.MustCompile(`(\[([a-zA-Z]*)\])`)
+	return sampleRegexp.ReplaceAllString(fieldVal, ".$2")
+}
+
+func trimInvalidDollarSign(fieldVal string) string {
+	if idx := strings.Index(fieldVal, ".$"); idx != -1 {
```
hasheddan, line 181: Do we always expect `.$` leading in the path if present? If so, I believe this is functionally equivalent to [`TrimPrefix`](https://pkg.go.dev/strings#TrimPrefix), if not it looks like we are cutting off everything before the `.$`, which I believe would truncate our field path

hasheddan (commented):
@tnthornton awesome to see this! A few nitpicks and questions, but mostly looks ready to go to me!

internal/xpkg/snapshot/xrd.go
tnthornton: This should being doing the opposite of TrimPrefix. This handles the case where we have a path like so:
```
spec.versions[0].schema.openAPIV3Schema.$ref"
```
This function returns:
```
spec.versions[0].schema.openAPIV3Schema
```

The reason for that is the first string is not a valid yaml path. So I do what I can to get the error close to the problem point and highlight that they are using an invalid field.

FWIW, I also tried TrimSuffix, however the key with a `.$` prefix is not guaranteed to be `$ref`. The end user can type whatever which is why I ultimately went with the above approach.

hasheddan: Ah! My apologies, I thought I saw `fieldVal[idx:]`, which I would have caught if I looked closer at your unit test!

hasheddan (approved):

internal/xpkg/snapshot/xrd.go
tnthornton: Ha! No worries 👍

PR #149, opened 2022-02-03
Capture validators for XRD if schema def is broken

tnthornton:
The previous error handling in validators.go would result in the validator for the XRD not being registered if the `openAPIV3Schema` object was invalid. Instead the desired behavior is to note that it happened (log it) and move on to trying to register the XRD's validator.

I have:

Verified that diagnostics would render again for a given XRD when an invalid schema definition was presented.
Verified that a warn diagnostic was not returned for the given XRD when an invalid schema definition was presented.

### DIFF
--- a/internal/xpkg/snapshot/validators.go
+++ b/internal/xpkg/snapshot/validators.go
@@ -39,9 +39,10 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


hasheddan (approved):
LGTM!

PR #163, opened 2022-03-29
Add `up xpkg xp-extract` command as EXPERIMENTAL

hasheddan:

Adds an xp-extract command that extracts xpkg contents and saves them to
a Crossplane cache compatible format. The logic in this command is
almost identical to that in the Crossplane package manager.

> NOTE: this enables folks who want to pre-populate the Crossplane package cache to have a similar experience as prior to Crossplane v1.7. However, we add this command as EXPERIMENTAL because Crossplane does not make any promises about backwards compatibility for its package cache format.

I have:

`up xpkg xp-extract crossplane/provider-aws:v0.24.1` and verified that `out.gz` contained the correct package contents (decompressed using `gunzip`). Then started Crossplane `v1.7.0` with it in the package cache and used it as a [`Provider` package source](https://doc.crds.dev/github.com/crossplane/crossplane/pkg.crossplane.io/Provider/v1@v1.7.0#spec-package) with `packagePullPolicy: Never`. Verified that `provider-aws:v0.24.1` was successfully installed.

### DIFF
--- a/cmd/up/xpkg/xpextract.go
+++ b/cmd/up/xpkg/xpextract.go
@@ -17,16 +17,17 @@ package xpkg
 import (
 	"archive/tar"
 	"compress/gzip"
+	"context"
 	"io"
 	"path/filepath"
 
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/go-containerregistry/pkg/v1/mutate"
-	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/pkg/errors"
 	"github.com/spf13/afero"
 
 	"github.com/upbound/up/internal/xpkg"
+	"github.com/upbound/up/internal/xpkg/dep/resolver/image"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


cmd/up/xpkg/xpkg.go
```diff
@@ -16,8 +16,9 @@ package xpkg
 
 // Cmd contains commands for interacting with xpkgs.
 type Cmd struct {
-	Build buildCmd `cmd:"" group:"xpkg" help:"Build a package."`
-	Init  initCmd  `cmd:"" group:"xpkg" help:"Initialize a package."`
-	Dep   depCmd   `cmd:"" group:"xpkg" help:"Manage package dependencies."`
-	Push  pushCmd  `cmd:"" group:"xpkg" help:"Push a package."`
+	Build     buildCmd     `cmd:"" group:"xpkg" help:"Build a package."`
+	XPExtract xpExtractCmd `cmd:"" group:"xpkg" help:"[EXPERIMENTAL] Extract package contents into a Crossplane cache compatible format."`
```
tnthornton, line 20: Do we see potentially supporting other extraction formats? I'm wondering if `up xpkg extract --format=xp` or something like that, would provide more flexibility in that case - e.g. if we introduce an `upbound` format would we also add a `upbound-extract` subcommand?

cmd/up/xpkg/xpextract.go
```diff
@@ -0,0 +1,157 @@
...
+
+	Tag    string `arg:"" help:"Tag of the package to extract. Must be a valid OCI image tag."`
+	Output string `short:"o" help:"Package output file path. Extension must be .gz or will be replaced." default:"out.gz"`
+}
+
+// Run runs the xp extract cmd.
+func (c *xpExtractCmd) Run() error { //nolint:gocyclo
+	// NOTE(hasheddan): most of the logic in this method is from the machinery
+	// used in Crossplane's package cache and should be updated to use shared
+	// libraries if moved to crossplane-runtime.
```
tnthornton, line 71: 👍

cmd/up/xpkg/xpkg.go
hasheddan: Good question! My primary reservation around that is some of the discussions of having an `up xpkg extract` command in the future that allows for users to reproduce the source code representation from a package, which would be a pretty fundamentally different operation. I intentionally went with a somewhat clunky `xp-extract` name as the experimental option 😅 What do you think?

tnthornton: I see. Let's just plan to revisit all of this before moving this subcommand out of experimental 👍 .

tnthornton (approved):

tnthornton (approved):

PR #168, opened 2022-04-20
Add up Docker credential helper

hasheddan:

Adds a new binary for docker-credential-up, which supplies credentials
to OCI clients in accordance with the docker credential helper protocol.

See other example implementations here:
- https://github.com/awslabs/amazon-ecr-credential-helper
- https://github.com/GoogleCloudPlatform/docker-credential-gcr

I have:

Tested locally with the following Docker config and verified successful push and pull to private repositories, which were unable to be pulled without:
```
{
  "credHelpers": {
    "<dev-url>": "up"
  }
}

```


Makefile
```diff
@@ -28,7 +28,7 @@ NPROCS ?= 1
 # to half the number of CPU cores.
 GO_TEST_PARALLEL := $(shell echo $$(( $(NPROCS) / 2 )))
 
-GO_STATIC_PACKAGES = $(GO_PROJECT)/cmd/up
+GO_STATIC_PACKAGES = $(GO_PROJECT)/cmd/up $(GO_PROJECT)/cmd/docker-credential-up
```
hasheddan, line 31: A few more updates will be needed in the `Makefile` here to make sure we are publishing this separately from the `up` binary.

cmd/docker-credential-up/main.go
```diff
@@ -0,0 +1,43 @@
...
+	"fmt"
+	"os"
+
+	"github.com/docker/docker-credential-helpers/credentials"
+
+	"github.com/upbound/up/internal/credhelper"
+	"github.com/upbound/up/internal/version"
+)
+
+const profile = "UP_PROFILE"
```
tnthornton, line 28: Not really sure what to do about this. We have env tags for "UP_PROFILE" defined in a few subcommands.  It would be nice if we could define this in one place and reference it in the others, but I don't know if that's something we can do with Kong.

internal/config/source_test.go
```diff
@@ -24,9 +24,59 @@ import (
...
+			reason: "If path is supplied but doesn't already exist create it.",
+			modifiers: []FSSourceModifier{
+				func(f *FSSource) {
+					f.path = "/.up/config.json"
+					f.fs = afero.NewMemMapFs()
+				},
+			},
+		},
+		"SuccessSuppliedExists": {
+			reason: "If path is supplied an already exists do not return error.",
```
tnthornton, line 57:
```suggestion
			reason: "If path is supplied and already exists do not return error.",
```

tnthornton (approved):

cmd/docker-credential-up/main.go
hasheddan: Good call out 👍🏻 One thing we can do for anywhere that actually uses Kong is define a common options package that all commands consume. I believe you can also build commands dynamically, so might be something to look at as well.

PR #169, opened 2022-04-21
Clear token from config if logout successful

hasheddan:

Updates the logout command to remove the token from the specified config
profile after it is successfully invalidated. This makes it obvious that
a token is no longer valid, and allows for surfacing more accurate
errors if a user attempts to access resources after logout.

I have:

Tested this locally by running `up login` and verifying that config is populated:
```
🤖 (up) cat ~/.up/config.json | jq .
{
  "upbound": {
    "default": "default",
    "profiles": {
      "default": {
        "id": "hasheddan",
        "type": "user",
        "session": "<token-omitted>",
        "account": "hasheddan"
      }
    }
  }
}
```

Then verified that running `up logout` causes the token to be removed:

```
🤖 (up) cat ~/.up/config.json | jq .
{
  "upbound": {
    "default": "default",
    "profiles": {
      "default": {
        "id": "hasheddan",
        "type": "user",
        "account": "hasheddan"
      }
    }
  }
}
```

### DIFF
--- a/cmd/up/logout.go
+++ b/cmd/up/logout.go
@@ -19,6 +19,7 @@ import (
 	"net/http"
 	"net/url"
 
+	"github.com/alecthomas/kong"
 	"github.com/pkg/errors"
 
 	"github.com/upbound/up-sdk-go"
@@ -30,11 +31,13 @@ import (
 const (
 	logoutPath = "/v1/logout"
 
-	errLogoutFailed = "unable to logout"
+	errLogoutFailed      = "unable to logout"
+	errGetProfile        = "failed to get profile"
+	errRemoveTokenFailed = "failed to remove token"
 )
 
 // AfterApply sets default values in login after assignment and validation.
-func (c *logoutCmd) AfterApply() error {
+func (c *logoutCmd) AfterApply(kongCtx *kong.Context) error {
 	src := config.NewFSSource()
 	if err := src.Initialize(); err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


nullable-eth (approved):

cmd/up/logout.go
```diff
@@ -43,18 +46,37 @@ func (c *logoutCmd) AfterApply() error {
...
 		}
 	}
+	// If account has not already been set, use the profile default.
+	if upCtx.Account == "" {
+		upCtx.Account = profile.Account
+	}
+	// If no account is set in profile, return an error.
+	if upCtx.Account == "" {
+		return errors.New(errNoAccount)
+	}
```
tnthornton, line 78: Do we set the account in the given profile when we perform a login operation? I'm curious how we might get into the state we're guarding against - I guess maybe if someone performs an `up logout` without ever performing an `up login`.

cmd/up/logout.go
```diff
@@ -74,13 +96,25 @@ type logoutCmd struct {
...
+		return errors.Wrap(err, errLogoutFailed)
+	}
+
+	// Logout is successful, remove token from config and update.
+	profile, err := upCtx.Cfg.GetUpboundProfile(upCtx.Profile)
+	if err != nil {
+		return errors.Wrap(err, errGetProfile)
+	}
+	profile.Session = ""
+	if err := upCtx.Cfg.AddOrUpdateUpboundProfile(upCtx.Profile, profile); err != nil {
```
tnthornton, line 116: Does this roughly map to an `upsert` operation? Out of scope for this PR, but wondering if `UpsertUpboundProfile` or `AddOrUpdateUpboundProfile` "rolls off the tongue" better.

tnthornton (approved):

cmd/up/logout.go
hasheddan: yeah -- I don't love everything about the config interface today, so I think it could use some love. I do like how the config has an underlying source, but the actual methods for interaction with the config and how they map to reading / writing the source could be improved 👍🏻

cmd/up/logout.go
hasheddan: Yeah we don't strictly _have_ to do this here, but I kept it in to keep the `upbound.Context` fully populated if we end up needing to consume it in `Run()` in the future.

PR #182, opened 2022-05-26
Copying vale fixes from upbound docs PR #215

plumbis:
Applying vale fixes from docs PR #215


hasheddan (commented):
@plumbis did this get updated with final revisions?

PR #197, opened 2022-07-14
Add support for experimental MCP API and drop control plane tokens

hasheddan:

Control plane tokens are now deprecated as self-hosted control planes
are no longer supported.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com

Adds an experimental MCP flag which allows for running control plane
commands against new MCP endpoints.

I have:

Testing in progress -- PR body will be updated with results.

=== UPDATE ===

Exercised existing commands and verified working correctly. Also demonstrated that new commands are invoked with `--mcp-experimental` (resulting in a failure with existing API).

```
🤖 (up) up ctp -h
Usage: up controlplane <command>

Interact with control planes.

Flags:
  -h, --help                         Show context-sensitive help.
  -v, --version                      Print version and exit.

      --mcp-experimental             Use experimental managed control planes API ($UP_MCP_EXPERIMENTAL).
      --domain=https://upbound.io    Root Upbound domain ($UP_DOMAIN).
      --profile=STRING               Profile used to execute command ($UP_PROFILE).
  -a, --account=STRING               Account used to execute command ($UP_ACCOUNT).

controlplane
  controlplane create <name>
    Create a hosted control plane.

  controlplane delete <id>
    Delete a control plane.

  controlplane list
    List control planes for the account.

  controlplane kubeconfig get --token=STRING <control-plane-ID>
    Get a kubeconfig for a control plane.
🤖 (up) up ctp list -a dan
NAME   ID                                     STATUS
luop   525b9759-82f9-4daf-8c1d-f9e58e9e9aef   ready
🤖 (up) up ctp create -a dan hello-world
🤖 (up) up ctp list -a dan
NAME          ID                                     STATUS
luop          525b9759-82f9-4daf-8c1d-f9e58e9e9aef   ready
hello-world   d26658ef-b04f-414c-bb39-b45c2697a63e   provisioning
🤖 (up) up ctp delete -a dan 525b9759-82f9-4daf-8c1d-f9e58e9e9aef
🤖 (up) up ctp list -a dan
NAME          ID                                     STATUS
luop          525b9759-82f9-4daf-8c1d-f9e58e9e9aef   deleting
hello-world   d26658ef-b04f-414c-bb39-b45c2697a63e   provisioning
🤖 (up) up ctp delete hello-world
Usage: up controlplane delete <id>

Delete a control plane.

Arguments:
  <id>    ID of control plane. ID is name if using experimental MCP API.

Flags:
  -h, --help                         Show context-sensitive help.
  -v, --version                      Print version and exit.

      --mcp-experimental             Use experimental managed control planes API ($UP_MCP_EXPERIMENTAL).
      --domain=https://upbound.io    Root Upbound domain ($UP_DOMAIN).
      --profile=STRING               Profile used to execute command ($UP_PROFILE).
  -a, --account=STRING               Account used to execute command ($UP_ACCOUNT).

up: error: invalid UUID length: 11
🤖 (up) up ctp delete hello-world --mcp-experimental
up: error: resource not found: {"message":"Not Found"}
```


cmd/up/controlplane/delete.go
```diff
@@ -20,15 +20,33 @@ import (
...
-	ID uuid.UUID `arg:"" required:"" help:"ID of control plane."`
+	id uuid.UUID
+
+	ID string `arg:"" help:"ID of control plane. ID is name if using experimental MCP API."`
 }
 
 // Run executes the delete command.
-func (c *DeleteCmd) Run(client *cp.Client) error {
-	return client.Delete(context.Background(), c.ID)
+func (c *DeleteCmd) Run(experimental bool, cc *cp.Client, oc *op.Client, upCtx *upbound.Context) error {
```
tnthornton, line 47: I wonder if it's worth adding experimental to the context versus passing it as a parameter 🤔.

Or maybe even better, we add functionality to group feature flags together that we can reference consistently.

Non-blocking for now, but I can see how if we keep adding feature flags that this will get kind of rough.

tnthornton (approved):

cmd/up/controlplane/delete.go
hasheddan: > I wonder if it's worth adding experimental to the context versus passing it as a parameter thinking.

@tnthornton I'm assuming you mean either the kong context or Upbound context here? Either way we would still be receiving it as a parameter, it would just be part of a struct, which as you point out I think could be useful (this is actually sort of what we are doing anyway by binding the bool to the kong context already) if we had multiple feature flags (it would become strictly necessary anyway if we had more than one `bool`). The reasoning for not going ahead and adding this to the Upbound context is that it gets passed around to commands where `--mcp-experimental` is not included, so this keeps it a little more tightly scoped.

Overall I totally agree that we need a better config story. For example, I'd love to be able to associate specific default env vars to each profile via the `config.json` 👍🏻 .

tnthornton: >(this is actually sort of what we are doing anyway by binding the bool to the kong context already) 

I meant more both adding and retrieving from the context directly. `context.WithValue` and `ctx.Value()`. Versus having it get passed by Kong.

> if we had multiple feature flags (it would become strictly necessary anyway if we had more than one bool). The reasoning for not going ahead and adding this to the Upbound context is that it gets passed around to commands where --mcp-experimental is not included, so this keeps it a little more tightly scoped.

Yep that all makes sense. I'm mostly just highlighting that we're shifting more and more towards needing a scalable solution as we continue to apply this approach of using feature flags 👍.

hasheddan: > I meant more both adding and retrieving from the context directly. context.WithValue and ctx.Value(). Versus having it get passed by Kong.

Definitely worth exploring though kong doesn't actually pass a `context.Context` to commands by default, so we would either need to bind a `context.Context` to the `kong.Context`, at which point it may just make sense to bind the value we want to pass itself -- anyway, we can experiment with a few approaches and see what feels clean 👍🏻

PR #230, opened 2022-08-15
Simplify `up upbound install` and add config command group

tnthornton:
### Description of your changes
This changeset seeks to address complexities in the current `up upbound install` process around the uninformative UX and the number of configuration steps end users are asked to go through to simply get started.

We address these items in by introducing the following:
* Adding progress messages to `up upbound install`
* Adding progress messages to `up upbound upgrade`
* Introduce `up config` with the following subcommands:
    * current
    * list
    * set
    * unset
* Introduce a standard `selfhosted` profile
* At the end of the `up upbound install`:
    * ExternalIP of the `ingress-nginx-controller` `Service` is outputted to the end user along with hostnames to add to their /etc/hosts file
    * The `selfhosted` profile and base config are updated and set as the default profile for the user in their up config.

Fixes xref: https://github.com/upbound/squad-orchestration/issues/26

I have:

- [X] Read and followed Upbound's [contribution process](https://git.io/fj2m9).
- [X] Run `make reviewable` to ensure this PR is ready for review.

#### `up upbound install --pretty`
https://user-images.githubusercontent.com/2375126/184567230-951f39f1-ad75-4575-b4a9-df8f28fd8bf4.mov

#### `up upbound install`
https://user-images.githubusercontent.com/2375126/184567336-99027e42-27f1-4a83-ab82-294f469ec330.mov

#### `up upbound upgrade --pretty`
https://user-images.githubusercontent.com/2375126/184567296-0348afb6-620e-4b56-94b6-87508664946a.mov

#### `up config` basics
```bash
# --help

up config --help
Usage: up config <command>

Interact with Upbound Profiles

Flags:
  -h, --help                         Show context-sensitive help.
  -v, --version                      Print version and exit.
  -q, --quiet                        Suppress all output.
      --pretty                       Pretty print output.

      --domain=https://upbound.io    Root Upbound domain ($UP_DOMAIN).
      --profile=STRING               Profile used to execute command ($UP_PROFILE).
  -a, --account=STRING               Account used to execute command ($UP_ACCOUNT).
      --mcp-experimental             Use experimental managed control planes API ($UP_MCP_EXPERIMENTAL).
      --insecure-skip-tls-verify     [INSECURE] Skip verifying TLS certificates ($UP_INSECURE_SKIP_TLS_VERIFY).

config
  config current
    Get current Upbound Profile.

  config list
    List Upbound Profiles.

  config set [<key> [<value>]]
    Set base configuration key, value pair in the Upbound Profile.

[...]

PR #231, opened 2022-08-16
Add support for organization management commands

### LINKED ISSUE #223: Support organization management commands

`up` does not currently allow a user to manage organizations.

In the immediate term, `up` should support:
- `up organization create`
- `up organization delete`
- `up organization list`
- `up organization update`

hasheddan:

Adds support for organization management commands:
- `up org create <name>`
- `up org delete <name>`
- `up org list`

> Note: we are opting not to support an `up org update` command for now as currently only the display name can be changed and it is unclear whether that will remain the case in the future.

Fixes #223 

I have:

```
🤖 (up) up org list
NAME                 ROLE  
upbound              owner 
crossplane           owner 
dan                  owner 
...
🤖 (up) up org delete dan
dan deleted
```

### DIFF
--- a/cmd/up/organization/list.go
+++ b/cmd/up/organization/list.go
@@ -1,0 +1,53 @@
+// Copyright 2022 Upbound Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package organization
+
+import (
+	"context"
+
+	"github.com/alecthomas/kong"
+	"github.com/pterm/pterm"
+
+	"github.com/upbound/up-sdk-go/service/organizations"
+
+	"github.com/upbound/up/internal/upbound"
+)
+
+// AfterApply sets default values in command after assignment and validation.
+func (c *listCmd) AfterApply(kongCtx *kong.Context, upCtx *upbound.Context) error {
+	kongCtx.Bind(pterm.DefaultTable.WithWriter(kongCtx.Stdout).WithSeparator("   "))
+	return nil
+}
+
+// listCmd lists organizations on Upbound.
+type listCmd struct{}
+
+// Run executes the list command.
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -235,6 +235,8 @@ replace (
 	github.com/golang/tools => ./internal/vendor/golang.org/x/tools
 	github.com/googleapis/gnostic => github.com/google/gnostic v0.5.5
 
+	github.com/upbound/up-sdk-go => github.com/hasheddan/up-sdk-go-1 v0.1.1-0.20220816115556-167771c88d9b
+
```
hasheddan, line 239: Will drop this after https://github.com/upbound/up-sdk-go/pull/30 is merged.

cmd/up/organization/organization.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		return err
+	}
+	cfg, err := upCtx.BuildSDKConfig(upCtx.Profile.Session)
+	if err != nil {
+		return err
+	}
+	kongCtx.Bind(upCtx)
+	kongCtx.Bind(organizations.NewClient(cfg))
+	kongCtx.Bind(organizations.NewClient(cfg))
```
epk, line 38:
```suggestion
	kongCtx.Bind(upCtx)
	kongCtx.Bind(organizations.NewClient(cfg))
```

cmd/up/organization/list.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+// AfterApply sets default values in command after assignment and validation.
+func (c *listCmd) AfterApply(kongCtx *kong.Context, upCtx *upbound.Context) error {
+	kongCtx.Bind(pterm.DefaultTable.WithWriter(kongCtx.Stdout).WithSeparator("   "))
+	return nil
+}
+
+// listCmd lists organizations on Upbound.
+type listCmd struct{}
```
epk, line 35: Nit
```suggestion
// listCmd lists organizations on Upbound.
type listCmd struct{}

// AfterApply sets default values in command after assignment and validation.
func (c *listCmd) AfterApply(kongCtx *kong.Context, upCtx *upbound.Context) error {
	kongCtx.Bind(pterm.DefaultTable.WithWriter(kongCtx.Stdout).WithSeparator("   "))
	return nil
}
```

cmd/up/organization/organization.go
hasheddan: Ah good catch!

epk (approved):

cmd/up/organization/list.go
hasheddan: I think this is a reasonable nit here (i.e. define the struct before the methods), though we have employed a pattern of putting the pre-run hooks at the top of files, so I think I'd prefer to move them all at once if we decide to do so 👍🏻

hasheddan:
Holding this one for #230 so that merge conflicts can be handled as part of this smaller diff

Issue #215, opened 2022-08-05
Support robot management commands


hasheddan:
### What problem are you facing?

Organization admins can manage robot accounts and tokens for them. However, creating / listing / updating / deleting robots and tokens is not currently supported by the CLI. 

### How could Upbound help solve your problem?

`up` should support the following commands for robot management:
- `up robot list`
- `up robot create`
- `up robot update`
- `up robot delete`
- `up robot token list <robot-id>`
- `up robot token create <robot-id>`
- `up robot token delete <robot-id>`

PR #234, opened 2022-08-18
Support robot management commands

hasheddan:

Adds commands to support create / delete / list of robots.

Adds commands for create / delete / list of robots.

Fixes https://github.com/upbound/up/issues/215

I have:

```
🤖 (up) up robot list -a dan
NAME       ID                                     DESCRIPTION   CREATED
cool-bot   baf29497-bf22-4b0d-81af-7ee692c474d8                 7h27m  
🤖 (up) up robot delete cool-bot
up: error: robots are not currently supported for user accounts
🤖 (up) up robot delete cool-bot -a dan
dan/cool-bot deleted
🤖 (up) up robot create super-cool-bot -a dan
dan/super-cool-bot created
🤖 (up) up robot list -a dan
NAME             ID                                     DESCRIPTION   CREATED
super-cool-bot   c06bfc00-acda-4407-affa-357b669faf7d                 6s     
🤖 (up) up robot token list super-cool-bot -a dan
No tokens found for robot super-cool-bot in dan
🤖 (up) up robot token create super-cool-bot cool-token -a dan
dan/super-cool-bot/cool-token created
Refusing to emit sensitive output. Please specify output location.
🤖 (up) up robot token create super-cool-bot super-cool-token -a dan -o -
dan/super-cool-bot/super-cool-token created

Access ID: REDACTED
Token: REDACTED
🤖 (up) up robot token create super-cool-bot super-cool-token-2 -a dan -o creds.json
dan/super-cool-bot/super-cool-token-2 created
🤖 (up) cat creds.json
{"accessId":"REDACTED","token":"REDACTED"}
```

### DIFF
--- a/cmd/up/robot/delete.go
+++ b/cmd/up/robot/delete.go
@@ -60,22 +60,23 @@ func (c *deleteCmd) Run(p pterm.TextPrinter, ac *accounts.Client, oc *organizati
 	// must guarantee that exactly one robot exists in the specified account
 	// with the provided name. Logic should be simplified when the API is
 	// updated.
-	var id uuid.UUID
-	found := false
+	var id *uuid.UUID
 	for _, r := range rs {
 		if r.Name == c.Name {
-			if found && !c.Force {
+			if id != nil && !c.Force {
 				return errors.Errorf(errMultipleRobotFmt, c.Name, upCtx.Account)
 			}
-			id = r.ID
-			found = true
+			// Pin range variable so that we can take address.
+			r := r
+			id = &r.ID
 		}
 	}
-	if !found {
+
+	if id == nil {
 		return errors.Errorf(errFindRobotFmt, c.Name, upCtx.Account)
 	}
 
-	if err := rc.Delete(context.Background(), id); err != nil {
+	if err := rc.Delete(context.Background(), *id); err != nil {
 		return err
 	}
 	p.Printfln("%s/%s deleted", upCtx.Account, c.Name)
--- a/cmd/up/robot/delete.go
+++ b/cmd/up/robot/delete.go
@@ -1,0 +1,83 @@
+// Copyright 2022 Upbound Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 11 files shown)


cmd/up/robot/delete.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		return err
+	}
+	if len(rs) == 0 {
+		return errors.Errorf(errFindRobotFmt, c.Name, upCtx.Account)
+	}
+	// TODO(hasheddan): because this API does not guarantee name uniqueness, we
+	// must guarantee that exactly one robot exists in the specified account
+	// with the provided name. Logic should be simplified when the API is
+	// updated.
```
hasheddan, line 62: @tnthornton @epk I think y'all both have the context on this situation. One could say it is valid to not check for duplicates here, but I would prefer that we opt to eliminate the possibility of a user being surprised until the behavior of this API actually guarantees safety (i.e. I am willing to pay this performance penalty in the context of a CLI tool to better protect the user). Both `delete` commands also have a "break glass" `--force` flag if necessary (hidden by default as this will be able to be removed in the future and I don't want folks to rely on it).

tnthornton (approved):

cmd/up/robot/delete.go
```diff
@@ -0,0 +1,83 @@
...
+			found = true
+		}
+	}
+	if !found {
+		return errors.Errorf(errFindRobotFmt, c.Name, upCtx.Account)
+	}
+
+	if err := rc.Delete(context.Background(), id); err != nil {
+		return err
+	}
```
epk, line 80: Nit: 
```suggestion
	var id *uuid.UUID
	for _, r := range rs {
		if r.Name == c.Name {
			if id != nil && !c.Force {
				return errors.Errorf(errMultipleRobotFmt, c.Name, upCtx.Account)
			}
			id = &r.ID
			break
		}
	}
	if id == nil {
		return errors.Errorf(errFindRobotFmt, c.Name, upCtx.Account)
	}

	if err := rc.Delete(context.Background(), &id); err != nil {
		return err
	}
```

hasheddan: @epk this will actually cause unexpected behavior here because we are taking the address of a range variable. In this case (when there is no conflict) `id` is always going to result in assigning to the last item in the list. We can still do what you suggest if we pin the range variable prior to taking the address.

Here is an example showing the problem and the solution: https://go.dev/play/p/baXvchHaQEe

I'll update to pin the range variable :) I had previously opted not to because we are using the `id` in a few places in some of these commands, but we can us separate vars 👍🏻

PR #238, opened 2022-08-19
Add --force option to destructive commands

### LINKED ISSUE #232: Prompt user for confirmation when performing destructive actions

When users are performing destructive actions, such as deleting a repository, control plane, or organization, we should prompt them for input to confirm they want to do so.

Update all potentially destructive commands to prompt for confirmation.

jastang:
Signed-off-by: Jason Tang <jason@upbound.io>

This PR adds a `--force` option to destructive commands, prompting the user by default to confirm the deletion of the entity (repository, robot, robot token, organization). If `--force` is supplied, no prompt is given.

If the user backs out of the delete, we will return an error, which will show the CLI usage by default.

I hope I know what I'm doing! ; )

Fixes #232 

I have:

Tested commands with and without the `--force` option on dev, except the `org` as I cannot create additional orgs.

### DIFF
--- a/cmd/up/repository/delete.go
+++ b/cmd/up/repository/delete.go
@@ -34,19 +34,21 @@ func (c *deleteCmd) BeforeApply() error {
 
 // AfterApply accepts user input by default to confirm the delete operation.
 func (c *deleteCmd) AfterApply(p pterm.TextPrinter, upCtx *upbound.Context) error {
-	if !c.Force {
-		confirm, err := c.prompter.Prompt("Are you sure you want to delete this repository? [y/n]", false)
-		if err != nil {
-			return err
-		}
-
-		if input.InputYes(confirm) {
-			p.Printfln("Deleting repository %s/%s. This cannot be undone.", upCtx.Account, c.Name)
-		} else {
-			return fmt.Errorf("operation canceled")
-		}
+	if c.Force {
+		return nil
 	}
-	return nil
+	confirm, err := c.prompter.Prompt("Are you sure you want to delete this repository? [y/n]", false)
+	if err != nil {
+		return err
+	}
+
+	if input.InputYes(confirm) {
+		p.Printfln("Deleting repository %s/%s. This cannot be undone.", upCtx.Account, c.Name)
+		return nil
+	}
+
+	return fmt.Errorf("operation canceled")
+
 }
 
 // deleteCmd deletes a repository on Upbound.
@@ -55,7 +57,7 @@ type deleteCmd struct {
 
 	Name string `arg:"" required:"" help:"Name of repository."`
 
-	Force bool `help:"Force deletion of repository." default:"false" short:"f"`
+	Force bool `help:"Force deletion of repository." default:"false"`
 }
 
 // Run executes the delete command.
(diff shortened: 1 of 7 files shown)


cmd/up/repository/delete.go
```diff
@@ -16,17 +16,46 @@ package repository
...
+			return err
+		}
+
+		if input.InputYes(confirm) {
+			p.Printfln("Deleting repository %s/%s. This cannot be undone.", upCtx.Account, c.Name)
+		} else {
+			return fmt.Errorf("operation canceled")
+		}
+	}
+	return nil
```
hasheddan, line 49: @jastang we typically try to return early to avoid doing extraneous work and the cognitive burden of nested branching (ref: https://github.com/crossplane/crossplane/blob/master/CONTRIBUTING.md#return-early)
```suggestion
	if !c.Force {
		return nil
	}
	confirm, err := c.prompter.Prompt("Are you sure you want to delete this repository? [y/n]", false)
	if err != nil {
		return err
	}

	if input.InputYes(confirm) {
		p.Printfln("Deleting repository %s/%s. This cannot be undone.", upCtx.Account, c.Name)
		return nil
	}
	return fmt.Errorf("operation canceled")
```

cmd/up/repository/delete.go
```diff
@@ -16,17 +16,46 @@ package repository
...
+}
+
 // deleteCmd deletes a repository on Upbound.
 type deleteCmd struct {
+	prompter input.Prompter
+
 	Name string `arg:"" required:"" help:"Name of repository."`
+
+	Force bool `help:"Force deletion of repository." default:"false" short:"f"`
```
hasheddan, line 58: This is definitely subjective, but since we use `-f` elsewhere for filepaths _and_ because `--force` is a destructive action, I would probably lean towards making the user type the full command out. What do you think?
```suggestion
	Force bool `help:"Force deletion of repository." default:"false"`
```

cmd/up/repository/delete.go
hasheddan: Side note: I was trying to decide whether the `operation canceled` scenario should return an error. In some ways it seems like a successful execution of what the user requested. I'm going to look up examples of this in other CLI tools and see how they handle, but did you have any references off the top of your head?

hasheddan (commented):
Thanks so much @jastang! Left a few comments below :)

cmd/up/repository/delete.go
jastang: Yeah, I agree with that and am not opinionated (just felt familiar). I can remove :)

cmd/up/repository/delete.go
hasheddan: > And is there an easy way to break out of the hook without executing Run()?

If you add a `*kong.Context` you can call `.Exit(0)` in the `AfterApply()`. Similarly, you can call `.PrintUsage()` if that is desirable as well. However, returning the error here might be just fine tbh :)

jastang: Thanks! I'll keep that in my back pocket. Just learning about Kong 🙂

jastang:
@hasheddan addressed your comments and re-tested. I think there's an implicit agreement to tackle `uninstall` separately -- I don't know how "atomic" those operations are, but just preferring not to invite any complexity I might not be able to handle 😅 

thanks for humouring me on this one!

hasheddan (approved):
Thanks @jastang!

PR #273, opened 2022-11-05
Fail build if sha256sum is not installed

AlainRoy:

This change will cause 'make' to fail if sha256sum is not installed.

Previously if sha256sum was not installed, make would print an error that was easily lost in the output, and the files containing the SHA sums were not created.

This does an upfront check for sha256sum and fails the build if it can't be found. 

I discovered this because sah256sum is not installed on Mac OS by default. 

Note: I could have included image.mk or imagelight.mk to define the binary. I didn't do that because I didn't want to pull all the other irrelevant definitions in those files. Also, this avoided modifying the build submodule and accidentally affecting other builds.

I have:

I ran 'make' multiple times, with and without sha256sum installed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,14 @@ GOLANGCILINT_VERSION = 1.47.3
 -include build/makelib/golang.mk
 
 # ====================================================================================
+# Setup binaries
+ifneq ($(shell type sha256sum 2>/dev/null),)
+SHA256SUM := sha256sum
+else
+$(error Please install sha256sum)
+endif
+
+# ====================================================================================
 # Targets
 
 # run `make help` to see the targets and options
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -36,6 +36,14 @@ GOLANGCILINT_VERSION = 1.47.3
 
 -include build/makelib/golang.mk
 
+# ====================================================================================
+# Setup binaries
+ifneq ($(shell type sha256sum 2>/dev/null),)
```
hasheddan, line 41: Would it make sense to fall back to shasum as we do in https://github.com/upbound/build/blob/dc23ed139e1eaf3df7cdf85c41433865a51ec489/makelib/imagelight.mk#L33

hasheddan (commented):
Thanks @AlainRoy ! One question below 🙂

Makefile
AlainRoy: I considered that, but I wanted to feel confident that I wouldn't break anything. This is a drop-in change that only adds error checking and doesn't change other behavior. 

Also, I see no clear reason to prefer shasum. It's a Perl script instead of a GNU utility. I'd prefer simplicity and just use the GNU utility. 

However, if you prefer this to look like the code in imagelight.mk, I can change it. Just let me know.

hasheddan: I’m good with how you have implemented 👍🏻 Thanks for the context!

hasheddan (approved):

PR #284, opened 2022-12-08
Add formatted ouput for more get & list commands

AlainRoy:
### Description
For the org, repo, robot, and robot commands, add support for formatted (json/yaml) output for the get and list commands.

Examples:

```
up org list --format json
[
    {
        "id": 1,
        "name": "upbound",
        "displayName": "Upbound Inc",
        "creatorId": 0,
        "role": "owner",
        "reservedEnvironments": 5
    },
    {
        "id": 829,
        "name": "upbound-dev",
        "displayName": "Upbound Dev",
        "creatorId": 0,
        "role": "owner",
        "reservedEnvironments": 0
    }
]
```

```
up org get upbound --format json
{
    "id": 1,
    "name": "upbound",
    "displayName": "Upbound Inc",
    "creatorId": 0,
    "role": "owner",
    "reservedEnvironments": 5
}
```

I have:

This code has been manually tested.

### DIFF
--- a/cmd/up/organization/list.go
+++ b/cmd/up/organization/list.go
@@ -23,6 +23,7 @@ import (
 	"github.com/upbound/up-sdk-go/service/organizations"
 
 	"github.com/upbound/up/internal/upbound"
+	"github.com/upbound/up/internal/upterm"
 )
 
 // AfterApply sets default values in command after assignment and validation.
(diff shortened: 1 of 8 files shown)


cmd/up/organization/list.go
```diff
@@ -44,15 +47,10 @@ func (c *listCmd) Run(p pterm.TextPrinter, pt *pterm.TablePrinter, oc *organizat
...
-// Prints a list of control planes. This is also used by the get command
-func printOrganizations(orgs []organizations.Organization, pt *pterm.TablePrinter) error {
-	data := make([][]string, len(orgs)+1)
-	data[0] = []string{"NAME", "ROLE"}
-	for i, o := range orgs {
-		data[i+1] = []string{o.Name, string(o.Role)}
-	}
-	return pt.WithHasHeader().WithData(data).Render()
+func extractFields(obj any) []string {
+	o := obj.(organizations.Organization)
```
hasheddan, line 54: nit: I don't think a change is necessarily needed, especially given that these `extractFields` funcs are quite local to their usage, but just noting that if type assertion fails here we will end up just printing the zero values of whatever fields are being extracted. We could opt to use more descriptive names, such as `extractOrganizationFields()`, but that doesn't actually give us additional type guarantees so I don't feel strongly. It would be cool to experiment with generics on the `printer.Print` calls in the future. Something like:

```
Print[T any](obj T, fieldNames []string, extractFields func(T) []string) error
```

Here is a more extensive runnable version if interested: https://go.dev/play/p/g0MxLhr6dc8

Anywho, none of this is blocking, just fun to discuss some future improvements :)

hasheddan (approved):
Thanks @AlainRoy !

PR #288, opened 2023-01-16
Add alpha commands for MCP connector

turkenh:

This PR adds the following two (alpha) commands to connect to an MCP from an application cluster:

- `up alpha ctp connect cp --token $API_TOKEN`: Creates a K8s secret with MCP Kubeconfig and installs MCP connector into the `kube-system` namespace (by default)
- `up alpha ctp bind storage.starter.org/v1alpha1`: Creates `APIService` for a given API version/group aggregating them from the connected MCP.

I have:

Tested against my MCP in production as a result of FRE.

```shell
❯ up alpha ctp list
NAME   ID                                     STATUS
cp     5b2e4d3e-58bf-4747-b91a-0337e8049aba   ready

❯ up alpha ctp connect cp --token $API_TOKEN
Connected to the MCP "cp", ready for binding APIs!

❯ up alpha ctp bind database.starter.org/v1alpha1
APIs under database.starter.org/v1alpha1 were bound to the Managed Control Plane!

❯ kubectl api-resources |grep starter
postgresqlinstances                            database.starter.org/v1alpha1          true         PostgreSQLInstance
```


cmd/up/controlplane/bind.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// Run executes the get command.
+func (c *bindCmd) Run(p pterm.TextPrinter, cc *cp.Client, upCtx *upbound.Context) error {
+	// Deploy APIService for the requested Group/Version
+	apiVersion := strings.Split(c.APIVersion, "/")
+	_, err := c.kAggregator.ApiregistrationV1().APIServices().Create(context.Background(), &apiregistrationv1.APIService{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: apiVersion[1] + "." + apiVersion[0],
+		},
```
tnthornton, line 65: If the end user (accidentally or deliberately) provides a malformed apiVersion string, I think this code will panic. For example (based on your example) if they input `up alpha ctp bind database.starter.org`.

I haven't tried this, but I wonder if we could use the `format:"X"` tag from https://github.com/alecthomas/kong#supported-tags to guard against that input.

cmd/up/controlplane/bind.go
```diff
@@ -0,0 +1,84 @@
...
+		Spec: apiregistrationv1.APIServiceSpec{
+			Group:   apiVersion[0],
+			Version: apiVersion[1],
+			Service: &apiregistrationv1.ServiceReference{
+				Namespace: c.Namespace,
+				Name:      "mcp-connector",
+			},
+			GroupPriorityMinimum:  1000,
+			VersionPriority:       15,
+			InsecureSkipTLSVerify: true,
```
tnthornton, line 75: Mostly just noting that we'll probably want to remove this in the future 👍. Is it worth adding a `TODO`? Maybe we could even add a flag that the end user specifies to override the default of `InsecureSkipTLSVerify` (in the future).

cmd/up/controlplane/connect.go
```diff
@@ -0,0 +1,199 @@
...
+
+// AfterApply sets default values in command after assignment and validation.
+func (c *connectCmd) AfterApply(kongCtx *kong.Context, upCtx *upbound.Context) error {
+	kubeconfig, err := kube.GetKubeConfig(c.Kubeconfig)
+	if err != nil {
+		return err
+	}
+
+	mgr, err := helm.NewManager(kubeconfig,
+		"mcp-connector",
```
tnthornton, line 58: We've referenced this same string in a couple of places in this package. Is it worth moving to a `const`?

tnthornton (commented):
Thanks @turkenh! The main comment I have is about the input validation for the `apiVersion` string; the rest of it is pretty superficial 👍

cmd/up/controlplane/bind.go
turkenh: There is an issue on the MCP connector side for tracking this, which needs to be resolved to get this dropped.
Before that, it does not make sense to make this configurable today, and we may consider it later if needed (e.g., once we default to `InsecureSkipTLSVerify: false`)
I added a todo here to indicate that.

cmd/up/controlplane/bind.go
turkenh: Good call, I just added some validation with a proper error message. 

Attempted to use `format:"X"` tag but couldn't really find a good documentation or example on how to use it and gave up given the check I've already added should be enough.

cmd/up/controlplane/bind.go
```diff
@@ -0,0 +1,101 @@
...
+
+	cp "github.com/upbound/up-sdk-go/service/controlplanes"
+	"github.com/upbound/up/internal/install"
+	"github.com/upbound/up/internal/kube"
+	"github.com/upbound/up/internal/upbound"
+)
+
+// AfterApply sets default values in command after assignment and validation.
+func (c *bindCmd) AfterApply(kongCtx *kong.Context, upCtx *upbound.Context) error {
+	kubeconfig, err := kube.GetKubeConfig("/Users/hasanturken/.kube/config")
```
hasheddan, line 39: @turkenh I'm guessing we don't want this to be hard-coded 🙂

cmd/up/controlplane/bind.go
```diff
@@ -0,0 +1,101 @@
...
+				"group":   apiVersion[0],
+				"version": apiVersion[1],
+				"service": map[string]interface{}{
+					"namespace": c.Namespace,
+					"name":      connectorName,
+				},
+				"groupPriorityMinimum": 1000,
+				"versionPriority":      15,
+				// TODO(turkenh): remove this when MCP connector has proper TLS setup
+				"insecureSkipTLSVerify": true,
```
hasheddan, line 90: Can we honor our existing global `--insecure-skip-tls-verify` flag here?

cmd/up/controlplane/connect.go
```diff
@@ -0,0 +1,201 @@
...
+
+	"github.com/upbound/up/internal/install"
+	"github.com/upbound/up/internal/install/helm"
+	"github.com/upbound/up/internal/kube"
+	"github.com/upbound/up/internal/upbound"
+)
+
+var (
+	mcpRepoURL = urlMustParse("https://charts.upbound.io/beta")
+)
```
hasheddan, line 43: TODO: move this and https://github.com/upbound/up/blob/f9db77ac27d35dbd122acbbfb70624dbe2b7dfc7/cmd/up/uxp/uxp.go#L31 to common location 👍🏻

cmd/up/controlplane/connect.go
```diff
@@ -0,0 +1,201 @@
...
+	_, err = c.mgr.GetCurrentVersion()
+	if err != nil {
+		return err
+	}
+
+	p.Printfln("Connected to the MCP %q, ready for binding APIs!", c.Name)
+	return nil
+}
+
+func (c *connectCmd) applyMCPKubeconfig(upCtx *upbound.Context) error {
```
hasheddan, line 128: TODO: @hasheddan address your TODOs! 😅 https://github.com/upbound/up/blob/a0614516bb1e31f85ea92047261594e4824e0c80/internal/kube/client.go#L50 (no action needed here for now)

cmd/up/controlplane/connect.go
```diff
@@ -0,0 +1,201 @@
...
+	params, err := c.parser.Parse()
+	if err != nil {
+		return errors.Wrap(err, errParseInstallParameters)
+	}
+	if err = c.mgr.Install("", params); err != nil {
+		return err
+	}
+
+	_, err = c.mgr.GetCurrentVersion()
+	if err != nil {
```
hasheddan, line 120: nitpick:
```suggestion
	if _, err := c.mgr.GetCurrentVersion(); err != nil {
```

cmd/up/controlplane/connect.go
```diff
@@ -0,0 +1,201 @@
...
+		Data: map[string][]byte{
+			"kubeconfig": b,
+		},
+	}, metav1.ApplyOptions{FieldManager: "up"})
+
+	if err != nil {
+		return errors.Wrap(err, "cannot apply MCP kubeconfig secret")
+	}
+
+	return nil
```
hasheddan, line 192: nitpick: `errors.Wrap` returns `nil` if `err == nil`
```suggestion
	return errors.Wrap(err, "cannot apply MCP kubeconfig secret")
```

hasheddan (commented):
@turkenh a few nitpicks that are mostly non-blocking, but I think we need to fix the hard-coded kubeconfig path prior to merge 👍🏻

cmd/up/controlplane/bind.go
turkenh: 🤦

cmd/up/controlplane/bind.go
turkenh: Well, not sure. 

I believe what this is for is something different than how that flag is being used. That said, the goal is to drop this soon: https://github.com/upbound/mcp-connector/issues/12

turkenh:
> @turkenh a few nitpicks that are mostly non-blocking, but I think we need to fix the hard-coded kubeconfig path prior to merge 👍🏻

Fixed it and responded to others.

cmd/up/controlplane/bind.go
hasheddan: Yeah this flag is typically used for indicating that `up` requests should skip TLS verify, but I actually feel okay about using it here to indicate that the generated kubeconfig should also skip. wdyt?

turkenh: We would have to use `--insecure-skip-tls-verify` in that case which would unrelatedly skip TLS verification for requests to Upbound API. 
This is about requests from the App cluster API server to MCP connector, which is totally unrelated.

I would rather not prefer to combine them, but if you feel strong about doing so, happy to change it.

hasheddan (approved):
LGTM @turkenh ! Might be worth cleaning up some of the go version bump / revert commits before merge 👍🏻

Issue #54, opened 2021-05-12
Support shell autocompletion


hasheddan:
`up` should support auto-completion for common shells, such as `bash`, `zsh`, `fish`, etc.

PR #290, opened 2023-02-23
Initial suport for tab completions

AlainRoy:
### Description

This adds initial support for tab completions. Specifically:

- A patched fork of the kongplete package. Kongplete makes it easy to add completions to kong, the command-line parsers. We're adding a forked package because it's missing support for command aliases. I [submitted a PR](https://github.com/WillAbides/kongplete/pull/17) about ten weeks ago, but it hasn't received any comments from the maintainer. 
- A new command, "install-completions" which provides a way to install the completions into your shell environment. 
- A completion predictor that allows us to complete the names of organizations. Once this PR is accepted, I'll extend it to support names for other objects, like control planes. I limited it to organizations for now to keep this PR small. 

Manual testing in bash. Kongplete also supports zsh and fish -- we'll verify those later. 

### Limitations
The way completion works with Kong and Kongplete means that when we do completion predictions, we don't have the results of parsing the command-line. This means that if someone specifies an option like --domain or --profile, we won't use it. Instead we'll use the default profile. For 99% of users, this will be fine. For internal developers switching between environments, they may find that tab completion doesn't use the profile they'd like to use. This is a hard problem to fix, but I think it's a reasonable starting place.


jastang:
Fixes #54

cmd/up/organization/organization.go
```diff
@@ -33,15 +37,46 @@ func (c *Cmd) AfterApply(kongCtx *kong.Context) error {
...
+	kongCtx.Bind(config.Helpers.OrgsClient)
 	return nil
 }
 
+func PredictOrgs() complete.Predictor {
+	return predictOrgs()
+}
+
+func predictOrgs() complete.PredictFunc {
+
```
hasheddan, line 51: Do we need these to be separate functions?
```suggestion
```

cmd/up/main.go
```diff
@@ -96,18 +98,19 @@ type cli struct {
...
+	Organization       organization.Cmd             `cmd:"" name:"organization" aliases:"org" help:"Interact with organizations."`
+	Profile            profile.Cmd                  `cmd:"" help:"Interact with Upbound profiles."`
+	Repository         repository.Cmd               `cmd:"" name:"repository" aliases:"repo" help:"Interact with repositories."`
+	Robot              robot.Cmd                    `cmd:"" name:"robot" help:"Interact with robots."`
+	Upbound            upbound.Cmd                  `cmd:"" maturity:"alpha" help:"Interact with Upbound."`
+	UXP                uxp.Cmd                      `cmd:"" help:"Interact with UXP."`
+	XPKG               xpkg.Cmd                     `cmd:"" help:"Interact with UXP packages."`
+	XPLS               xpls.Cmd                     `cmd:"" help:"Start xpls language server."`
+	Alpha              alpha                        `cmd:"" help:"Alpha features. Commands may be removed in future releases."`
+	InstallCompletions kongplete.InstallCompletions `cmd:"" help:"install shell completions"`
```
hasheddan, line 113: nitpick:
```suggestion
	InstallCompletions kongplete.InstallCompletions `cmd:"" help:"Install shell completions"`
```

internal/upbound/context.go
```diff
@@ -219,6 +226,31 @@ func (c *Context) BuildSDKConfig(session string) (*up.Config, error) {
 	}), nil
 }
 
+// MakeClients makes clients for use by completion predictors.
```
hasheddan, line 229: Is the idea here that we would build _all_ clients when we expand to completions for more commands? Most commands already construct clients, and I'm wondering if we could instead refactor those to work with their specific completion (e.g.  https://github.com/upbound/up/pull/290/files#diff-71435ee5337944477bebca87d5ae3fe509ade527c0edb8b31654797219a84176R31)

hasheddan (commented):
This is so awesome -- thanks @AlainRoy! Left a few questions -- was also wondering if it would be worth forking the library rather than vendoring it? Might avoid some churn if the PR gets accepted.

internal/upbound/context.go
AlainRoy: Yeah, that was what I was going for. I didn't see a clear way to refactor them, since the client I'm building can't use the command-line flags. I'll think about it a bit more.

cmd/up/organization/organization.go
AlainRoy: Dan -- I vendored instead of forking because that's what Taylor recommended. Given that the PR has been lingering for 10+ weeks, I don't think it's going to be accepted.

hasheddan: @AlainRoy ack -- I think I would lean towards forking and maintaining eventually if we want to consume this elsewhere (e.g. Crossplane), but this works for now!

hasheddan (approved):
Awesome -- thanks @AlainRoy!

PR #297, opened 2023-03-06
Fix linter errors and separate kubeconfig generation steps

muvaf:

I needed to separate the kubeconfig steps as I needed to construct a kube client but then one thing led to another and I had to do a big linter pass to make `make reviewable` pass. This included import ordering, some changes in the interfaces and addition of `context.Context` in places where it was deemed necessary by the `contextcheck` golinter.

I have:

Only meaningful change is made in kubeconfig generation and I tested that manually and confirmed that it's working in the same way.


internal/xpls/server/server.go
```diff
@@ -169,14 +169,15 @@ func (s *Server) DidChange(ctx context.Context, params *protocol.DidChangeTextDo
 		return
 	}
 
-	if err := s.snap.ReParseFile(filename); err != nil {
+	// nolint contextcheck because it seems like a linter error.
```
hasheddan, line 172: @muvaf can we pass a context here? (i.e. are we nolinting to get this to pass or do we not intend to ever pass a context?)

muvaf: I wasn't able to find the function in the tree that can accept `context.Context` but now I did and added it. Thanks!

hasheddan (approved):

PR #312, opened 2023-03-22
Added --public flag for creating configuration

AlainRoy:
Added --public flag for creating configuration

This is opposite of the API, which has a "private" flag. The idea is that we default to private (that is, default to "more secure"), but allow people to make it public if they want. 

I have:

Manual testing.


cmd/up/configuration/create.go
```diff
@@ -42,6 +42,7 @@ type createCmd struct {
 	Name       string `arg:"" required:"" help:"Name of configuration."`
 	Context    string `required:"" help:"Name of the GitHub account/org"`
 	TemplateId string `required:"" help:"Name of the configuration template" predictor:"templates"`
+	Public     bool   `default:"false" help:"Should Github repo should be created as public? (Default: false)"`
```
avalanche123, line 45: default `false` seems different from the UI flow, was this intentional?

AlainRoy: I described my reasoning above. I wasn't aware that the default in the UI would be public, but assumed we'd want to default to private repositories. 

If the UI is going to default to public repos, I can switch it. Although I think private repos make more sense, it's more important to match the Console.

AlainRoy: I discussed with Jason and will switch to public by default. I'll close this PR and do a new one.

Issue #321, opened 2023-04-05
package.yaml has empty controller image even if it is not given by user


muvaf:
### What happened?

I got an error deploying my provider built with `v0.16.0` : `...spec.containers[0].image: required value` which is surprising because it should be the package itself since I didn't give any package URL [in the repo](https://github.com/upbound/provider-dummy/blob/main/package/crossplane.yaml) similar to other providers. When I extracted `/package.yaml` from the package, I saw the following:
```yaml
---
apiVersion: meta.pkg.crossplane.io/v1alpha1
kind: Provider
metadata:
  ...
  creationTimestamp: null
  name: provider-dummy
spec:
  controller:
    image: ""
```

When I downgrade to `v0.15.0`, the problem disappears. I assume we made a change in between that defaulted `spec.controller.image` and `metadata.creationTimestamp` to their zero-values.

### How can we reproduce it?

Build a provider package with `v0.16.0`. Then run it with classic `docker run -d` and while it's running, extract `package.yaml` with `docker cp <container hash>:/package.yaml /tmp/package.yaml` and see the beginning of that file.

### What environment did it happen in?

up `v0.16.0`

muvaf:
Seems like we're still using `v1.5.0` of Crossplane that does not have `omitempty` json tag for the image field, hence initialize to `""`. Updating it causes dependency problems though. Will continue.

PR #324, opened 2023-04-05
go.mod: update to crossplane v1.6.0

muvaf:

`up` CLI is still using the old meta API type for `Provider` type which does not have `omitempty` for the `spec.controller` field and this causes it to be initialized with `""` and when Crossplane sees that, it tries to set the corresponding Deployment image to empty string and fails. I'm not entirely sure how we were getting away with this so far.

Fixes https://github.com/upbound/up/issues/321

I have:

Built a provider package and confirmed that when not given, `spec.controller` field does not appear in the built package.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,8 +5,8 @@ go 1.18
 require (
 	github.com/Masterminds/semver v1.5.0
 	github.com/alecthomas/kong v0.6.1
-	github.com/crossplane/crossplane v1.5.0
-	github.com/crossplane/crossplane-runtime v0.15.1-0.20210930095326-d5661210733b
+	github.com/crossplane/crossplane v1.6.0
+	github.com/crossplane/crossplane-runtime v0.15.1-0.20211029211307-c72bcdd922eb
 	github.com/crossplane/crossplane/controller/apiextensions v0.0.0-00010101000000-000000000000
 	github.com/crossplane/crossplane/xcrd v0.0.0-00010101000000-000000000000
 	github.com/docker/docker-credential-helpers v0.6.4
--- a/internal/xpkg/dep/cache/entry.go
+++ b/internal/xpkg/dep/cache/entry.go
@@ -20,15 +20,14 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/crossplane/crossplane-runtime/pkg/errors"
+	xpv1 "github.com/crossplane/crossplane/apis/apiextensions/v1"
 	"github.com/spf13/afero"
-
 	v1ext "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	v1beta1ext "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/yaml"
 
-	"github.com/crossplane/crossplane-runtime/pkg/errors"
-	xpv1 "github.com/crossplane/crossplane/apis/apiextensions/v1"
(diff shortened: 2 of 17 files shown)


internal/xpkg/lint.go
```diff
@@ -133,24 +132,16 @@ func IsValidatingWebhookConfiguration(o runtime.Object) error {
 
 // IsXRD checks that an object is a CompositeResourceDefinition.
 func IsXRD(o runtime.Object) error {
-	switch o.(type) {
-	case *v1beta1.CompositeResourceDefinition:
-		return nil
-	case *v1.CompositeResourceDefinition:
-		return nil
-	default:
+	if _, ok := o.(*v1.CompositeResourceDefinition); ok {
```
muvaf, line 135: `v1beta1` for `apiextensions` group does not exist anymore in v1.6.0 and after.

go.mod
```diff
@@ -5,8 +5,8 @@ go 1.18
 require (
 	github.com/Masterminds/semver v1.5.0
 	github.com/alecthomas/kong v0.6.1
-	github.com/crossplane/crossplane v1.5.0
-	github.com/crossplane/crossplane-runtime v0.15.1-0.20210930095326-d5661210733b
+	github.com/crossplane/crossplane v1.6.0
```
muvaf, line 8: The latest is `v1.11.3` but it requires too many changes that we wouldn't be able to cut a patch release.

jastang:
xref https://github.com/upbound/provider-aws/pull/647

internal/xpkg/dep/cache/entry.go
```diff
@@ -238,15 +235,9 @@ func (e *entry) writeObjects(objs []runtime.Object) (*flushstats, error) { // no
 		case *v1ext.CustomResourceDefinition:
 			name = crd.GetName()
 			inc = stats.incCRDs
-		case *v1beta1.CompositeResourceDefinition:
-			name = crd.GetName()
-			inc = stats.incXRDs
```
tnthornton, line 243: If we remove these, I don't think we'll be able to support folks using XRDs from this APIVersion. Similar with the below v1beta1.Composition being removed.

tnthornton (commented):
Most of the comments I have roll up into the one I left. We're pretty clearly removing support for v1beta1. I'm missing a little bit of context, is there a reason why we need to remove support for these versions?

muvaf:
> I'm missing a little bit of context, is there a reason why we need to remove support for these versions?

@tnthornton It's a side effect of updating to Crossplane v1.6.0 which [removed](https://github.com/crossplane/crossplane/pull/2804) the v1beta1 versions of those CRDs altogether.

tnthornton:
> @tnthornton It's a side effect of updating to Crossplane v1.6.0 which https://github.com/crossplane/crossplane/pull/2804 the v1beta1 versions of those CRDs altogether.

Gotcha. I doubt we have many/any users of xpls working with APIs from <1.6.0. Could we maybe put a note about minimum supported versions in the README? I think as a long as we communicate it, we can move forward with the current changeset 👍

jastang:
> > @tnthornton It's a side effect of updating to Crossplane v1.6.0 which [crossplane/crossplane#2804](https://github.com/crossplane/crossplane/pull/2804) the v1beta1 versions of those CRDs altogether.
> 
> Gotcha. I doubt we have many/any users of xpls working with APIs from <1.6.0. Could we maybe put a note about minimum supported versions in the README? I think as a long as we communicate it, we can move forward with the current changeset 👍

@tnthornton added the note in the README. @muvaf will be at kubecon this week so we could help cut the patch release 👍🏼

tnthornton (approved):

PR #330, opened 2023-04-26
Check validity of kubeconfig before applying it

### LINKED ISSUE #323: Get kubeconfig command does not fail and it is misleading

When you run `up ctp kubeconfig get` with wrong token, account or any other argument, it does  not fail as long as it's able to construct kubeconfig YAML. Then user gets error in `kubectl` when they try to access MCP like `api is not found` which is misleading.

`up ctp kubeconfig get` should check whether the kubeconfig it produces actually works before returning to user.

jastang:
Use a client to verify connectivity before applying a kubeconfig and setting it as the current context. Will not change the current context on a failure.

Fixes #323 

I have:

Bad token:
```
➜  bin ./up ctp kubeconfig get --token="fake token" my-ctp
up: error: the server has asked for the client to provide credentials
```

Good token, non-existent target:
```
./up ctp kubeconfig get --token=`cat my-token` does-not-exist
up: error: the server could not find the requested resource
```

Oops, wrong Upbound organization:
```
./up ctp kubeconfig get --token=`cat my-token` my-ctp     
up: error: the server could not find the requested resource
```

That's the one!
```
➜  bin ./up ctp kubeconfig get --token=`cat my-token` my-ctp -a upbound
Current context set to upbound-upbound-my-ctp
```

### DIFF
--- a/internal/kube/client.go
+++ b/internal/kube/client.go
@@ -20,7 +20,6 @@ import (
 	"path"
 	"strings"
 
-	"github.com/crossplane/crossplane-runtime/pkg/errors"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
@@ -38,10 +37,6 @@ const (
 
 	// UpboundK8sResource is appended to the end of the kubeconfig server path.
 	UpboundK8sResource = "k8s"
-)
-
-const (
-	errInvalidContext = "failed to discover crossplane api resources on control plane"
 )
 
 // GetKubeConfig constructs a Kubernetes REST config from the specified
@@ -73,7 +68,7 @@ func BuildControlPlaneKubeconfig(proxy *url.URL, id string, token string) *api.C
 
 // ApplyControlPlaneKubeconfig applies a control plane kubeconfig to an existing
 // kubeconfig file and sets it as the current context.
-func ApplyControlPlaneKubeconfigIfValid(mcpConf *api.Config, existingFilePath string) error {
+func ApplyControlPlaneKubeconfig(mcpConf *api.Config, existingFilePath string) error {
 	po := clientcmd.NewDefaultPathOptions()
 	po.LoadingRules.ExplicitPath = existingFilePath
 	conf, err := po.GetStartingConfig()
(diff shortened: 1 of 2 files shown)


internal/kube/client.go
```diff
@@ -67,7 +73,7 @@ func BuildControlPlaneKubeconfig(proxy *url.URL, id string, token string) *api.C
 
 // ApplyControlPlaneKubeconfig applies a control plane kubeconfig to an existing
 // kubeconfig file and sets it as the current context.
-func ApplyControlPlaneKubeconfig(mcpConf *api.Config, existingFilePath string) error {
+func ApplyControlPlaneKubeconfigIfValid(mcpConf *api.Config, existingFilePath string) error {
```
AlainRoy, line 76: Given that Go tends to have concise names, I would keep the original name of the function. It can already return an error, you're just extending the errors that can happen.

internal/kube/client.go
```diff
@@ -84,5 +90,37 @@ func ApplyControlPlaneKubeconfig(mcpConf *api.Config, existingFilePath string) e
...
+
+	if err != nil {
+		// For example, the target cluster does not exist.
+		return err
+	}
+
+	// This check is optional, but provides additional confidence that the control plane is healthy.
+	if coreTypes := len(resources.APIResources); coreTypes == 0 {
+		return errors.New(errInvalidContext)
+	}
```
AlainRoy, line 123: I'm uncertain about this check. What if someone tries to get a kubeconfig to debug a control plane that is in a bad state? What if they get the kubeconfig early, before Crossplane is fully set up?

I'm open to discussion, but my gut says that we probably shouldn't do this check.

AlainRoy:
I love what you've done! I have one nit I don't feel strongly about (name of the function), and one suggestion: I'm not sure you want the last check you've got. I'm open to discussion about it though.

internal/kube/client.go
jastang: I think that is a great point. I wasn't thrilled about it but I'm pretty convinced we don't want to unintentionally block anyone from debugging at this stage. I'll remove it!

AlainRoy (approved):
Looks good to me!

PR #332, opened 2023-05-09
xpkg batch subcommand to batch process provider packages

ulucinar:

This PR introduces an `xpkg batch` subcommand to batch process a family of provider packages. The `xpkg build` implementation dumps the built packages onto the file system for `xpkg push` to load them from the filesystem and push them to a package registry. These two subcommands should stay as is with this PR and a new `xpkg batch` command is added that does not need to store the built provider packages in the filesystem. The packages are instead directly pushed to the configured registry. `xpkg batch` has batch processing capabilities to streamline the build process of a family of smaller provider packages. An example invocation for the command looks like:
```
up xpkg batch --service monolith,config,acm,rds,ec2 --family-base-image build-e46f4576/provider-aws --provider-name provider-aws --family-package-url-format xpkg.upbound.io/upbound-release-candidates/%s:v0.35.0-rc.0 --package-repo-override monolith=provider-aws --package-repo-override config=provider-family-aws --provider-bin-root .../provider-aws/_output/bin --examples-root .../provider-aws/examples --examples-group-override monolith=* --examples-group-override config=providerconfig --auth-ext .../provider-aws/package/auth.yaml --crd-root .../provider-aws/package/crds --crd-group-override monolith=* --crd-group-override config=aws --package-metadata-template .../provider-aws/package/crossplane.yaml.tmpl --template-var XpkgRegOrg=xpkg.upbound.io/upbound-release-candidates --template-var DepConstraint=">= 0.0.0-0" --concurrency 4
```

All smaller packages built by this command share a common base image and this common base is specified with the `--family-base-image` option. The providers to be built and pushed are specified via the `--service` option. `xpkg batch` adds the `base` and the `upbound` layers as before, as well as a layer containing the provider binary. The `base` and `upbound` labeled layers are shared between all the architectures of a provider's package.

`xpkg batch` builds & pushes the provider packages in parallel with the configured (via the `--concurrency` option) maximum concurrency. 

`xpkg batch` also has a templating engine so that the family of providers being built can share a template provider metadata. The templating engine substitutes the `{{ .Service }}` and `{{ .Name }}` template variables with the service and package repo names, respectively, and further template variables can also be supplied via multiple `--template-var` command-line options. 

The `--family-package-url-format` should specify the package URL format string and must contain a single string specifier, which will be substituted with the repo name of the provider being processed. The provider binaries should reside under `--provider-bin-root` in their respective platform folders. The platforms for which the provider packages will be built can be specified via the `--platform` option. 

I have:

Tested both locally and in Github workflows against `upbound/provider-aws` packages.

### DIFF
--- a/cmd/up/xpkg/batch.go
+++ b/cmd/up/xpkg/batch.go
@@ -1,0 +1,399 @@
+// Copyright 2023 Upbound Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package xpkg
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"text/template"
+
+	"github.com/alecthomas/kong"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+
+	"github.com/upbound/up/internal/upbound"
+
+	"github.com/crossplane/crossplane-runtime/pkg/errors"
+	"github.com/crossplane/crossplane-runtime/pkg/parser"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/pterm/pterm"
+	"github.com/spf13/afero"
+
+	"github.com/upbound/up/internal/xpkg"
+	"github.com/upbound/up/internal/xpkg/parser/examples"
+	"github.com/upbound/up/internal/xpkg/parser/yaml"
+)
+
+const (
+	errInvalidTemplate    = "service-scoped provider template is not valid"
+	errMetadataBackend    = "failed to initialize the package metadata parser backend"
+	errCRDBackend         = "failed to initialize the CRD parser backend"
+	errTemplateFmt        = "failed to execute the provider metadata template using: %v"
+	errInvalidPlatformFmt = "failed to parse platform name. Expected syntax is <OS>_<arch>: %s"
+	errBuildPackageFmt    = "failed to service-scoped provider package: %s"
+	errGetConfigFmt       = "failed to get config file from %s image for service %q"
+	errMutateConfigFmt    = "failed to mutate config file from %s image for service %q"
+	errGetLayersFmt       = "failed to get layers from %s image for service %q"
+	errGetBaseLayersFmt   = "failed to get base layers from %s image for service %q"
+	errGetDigestFmt       = "failed to get layer's digest from %s image for service %q"
+	errAppendLayersFmt    = "failed to append layers to %s image for service %q"
+	errReadProviderBinFmt = "failed to read %q provider binary for %s platform from path: %s"
+	errNewLayerFmt        = "failed to initialize a new image layer for %s platform for service %q"
+	errAddLayerFmt        = "failed to add the service-scoped provider binary layer for %s platform for service %q"
+	errPushPackageFmt     = "failed to push service-scoped provider package: %s"
+)
+
+// AfterApply constructs and binds Upbound-specific context to any subcommands
+// that have Run() methods that receive it.
+func (c *batchCmd) AfterApply(kongCtx *kong.Context) error {
+	c.fs = afero.NewOsFs()
+	// NOTE(aru): we currently only support fetching family base image from
+	// daemon, but may opt to support additional sources in the future.
+	c.fetch = daemonFetch
+	if c.TemplateVar == nil {
+		c.TemplateVar = make(map[string]string)
+	}
+
+	upCtx, err := upbound.NewFromFlags(c.Flags)
+	if err != nil {
+		return err
+	}
+	kongCtx.Bind(upCtx)
+
+	return nil
+}
+
+// batchCmd builds and pushes a family of Crossplane provider packages.
+type batchCmd struct {
+	fs    afero.Fs
+	fetch fetchFn
+
+	FamilyBaseImage        string   `help:"Family image used as the base for the service-scoped provider packages." required:""`
+	ProviderName           string   `help:"Provider name, such as provider-aws to be used while formatting service-scoped provider package repositories." required:""`
+	FamilyPackageURLFormat string   `help:"Family package URL format to be used for the service-scoped provider packages. Must be a valid OCI image URL with the format specifier \"%s\", which will be substituted with <provider name>-<service name>." required:""`
+	Service                []string `help:"Services to build the scoped provider packages for." default:"monolith"`
+
+	Platform        []string `help:"Platforms to build the packages for. Each platform should use the <OS>_<arch> syntax. An example is: linux_arm64." default:"linux_amd64,linux_arm64"`
+	ProviderBinRoot string   `short:"p" help:"Provider binary paths root. Service-scoped provider binaries should reside under the platform directories in this folder." type:"existingdir"`
+
+	PackageMetadataTemplate string            `help:"Service-scoped provider metadata template. The template variables {{ .Service }} and {{ .Name }} will be substituted when the template is executed among with the supplied template variable substitutions." default:"./package/crossplane.yaml.tmpl" type:"path"`
+	TemplateVar             map[string]string `help:"Service-scoped provider metadata template variables to be used for the specified template."`
+
+	ExamplesGroupOverride map[string]string `help:"Overrides for the location of the example manifests folder of a service-scoped provider." optional:""`
+	CRDGroupOverride      map[string]string `help:"Overrides for the locations of the CRD folders of the service-scoped providers." optional:""`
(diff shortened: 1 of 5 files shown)


cmd/up/xpkg/batch.go
```diff
@@ -0,0 +1,481 @@
...
+		repo = fmt.Sprintf("%s-%s", c.ProviderName, s)
+	}
+	return repo
+}
+
+func (c *batchCmd) getPackageURL(s string) string {
+	return fmt.Sprintf(c.FamilyPackageURLFormat, c.getPackageRepo(s))
+}
+
+func getAddendumLayers(baseImg, img v1.Image, p, s string) ([]v1.Layer, [][2]string, error) {
```
turkenh, line 258: nit: using more explicit names for variables `p` and `s` would help for readability.

turkenh: Similarly using named result parameters would help https://go.dev/doc/effective_go#named-results

cmd/up/xpkg/batch.go
```diff
@@ -0,0 +1,481 @@
...
+}
+
+func (c *batchCmd) getBuilder(service string) (*xpkg.Builder, error) {
+	ex, err := filepath.Abs(c.getExamplesGroup(service))
+	if err != nil {
+		return nil, err
+	}
+
+	var authBE parser.Backend
+	if ax, err := filepath.Abs(c.AuthExt); err == nil {
```
turkenh, line 355: Should we handle `err != nil` ? (e.g. for debugging)

cmd/up/xpkg/batch.go
```diff
@@ -0,0 +1,481 @@
...
+	tries := c.PushRetry + 1
+	retryMsg := ""
+	for i := uint(0); i < tries; i++ {
+		p.Printfln("Pushing xpkg to %s.%s", t, retryMsg)
+		err := PushImages(p, upCtx, imgs, t, c.Create, c.Flags.Profile)
+		if err == nil {
+			break
+		}
+		if i == tries-1 { // no more retries
+			p.Printfln("Failed to push xpkg to %s.", t)
```
turkenh, line 238: Should we also print the error here?

cmd/up/xpkg/xpkg.go
```diff
@@ -32,4 +32,5 @@ type Cmd struct {
 	Init      initCmd      `cmd:"" help:"Initialize a package."`
 	Dep       depCmd       `cmd:"" help:"Manage package dependencies."`
 	Push      pushCmd      `cmd:"" help:"Push a package."`
+	Batch     batchCmd     `cmd:"" help:"Batch build and push a family of service-scoped provider packages."`
```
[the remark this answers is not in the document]
ulucinar: Done. Replaces occurrences of "service-scoped providers" with "smaller providers" in the help docs & error messages.

cmd/up/xpkg/batch.go
```diff
@@ -0,0 +1,481 @@
...
+
+// batchCmd builds and pushes a family of Crossplane provider packages.
+type batchCmd struct {
+	fs    afero.Fs
+	fetch fetchFn
+
+	FamilyBaseImage        string   `help:"Family image used as the base for the service-scoped provider packages." required:""`
+	ProviderName           string   `help:"Provider name, such as provider-aws to be used while formatting service-scoped provider package repositories." required:""`
+	FamilyPackageURLFormat string   `help:"Family package URL format to be used for the service-scoped provider packages. Must be a valid OCI image URL with the format specifier \"%s\", which will be substituted with <provider name>-<service name>." required:""`
+	Service                []string `help:"Services to build the scoped provider packages for." default:"monolith"`
```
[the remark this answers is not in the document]
ulucinar: Done.

cmd/up/xpkg/batch.go
ulucinar: Done. Also added some code comments to explain the method.

cmd/up/xpkg/batch.go
ulucinar: Done.

cmd/up/xpkg/batch.go
```diff
@@ -0,0 +1,481 @@
...
+			result = errors.Wrap(result, err.Error())
+		}
+	}
+	return errors.WithMessage(result, errBatch)
+}
+
+func (c *batchCmd) processService(p pterm.TextPrinter, upCtx *upbound.Context, baseImgMap map[string]v1.Image, s string) error { //nolint:gocyclo
+	imgs := make([]v1.Image, 0, len(c.Platform))
+	var addendumLayers []v1.Layer
+	var labels [][2]string
```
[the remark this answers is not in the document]
ulucinar: Done. Added comments to the function signature and also added some implementation comments.

cmd/up/xpkg/batch.go
ulucinar: Done.

turkenh (approved):
Thank you @ulucinar!

tnthornton (approved):

ulucinar:
@turkenh, could you please take a look at the last commit that introduces the option to be able to selectively dump provider packages onto the filesystem?

cmd/up/xpkg/xpkg.go
```diff
@@ -32,4 +32,5 @@ type Cmd struct {
 	Init      initCmd      `cmd:"" help:"Initialize a package."`
 	Dep       depCmd       `cmd:"" help:"Manage package dependencies."`
 	Push      pushCmd      `cmd:"" help:"Push a package."`
+	Batch     batchCmd     `cmd:"" help:"Batch build and push a family of service-scoped provider packages." hidden:""`
```
turkenh, line 35: Instead of adding this as `hidden`, we may want to introduce under alpha. 

```
up alpha xpkg --help
```

Did you consider that already?

cmd/up/xpkg/batch.go
```diff
@@ -0,0 +1,569 @@
...
+
+// batchCmd builds and pushes a family of Crossplane provider packages.
+type batchCmd struct {
+	fs    afero.Fs
+	fetch fetchFn
+
+	FamilyBaseImage        string   `help:"Family image used as the base for the smaller provider packages." required:""`
+	ProviderName           string   `help:"Provider name, such as provider-aws to be used while formatting smaller provider package repositories." required:""`
+	FamilyPackageURLFormat string   `help:"Family package URL format to be used for the smaller provider packages. Must be a valid OCI image URL with the format specifier \"%s\", which will be substituted with <provider name>-<service name>." required:""`
+	Services               []string `help:"Services to build the scoped provider packages for." default:"monolith"`
```
turkenh, line 99: Should we use another word instead of `Services`, e.g. `Subpackages`, with similar concerns as the previous comment as using _smaller providers_ instead of _service scoped_ ?

cmd/up/xpkg/batch.go
```diff
@@ -0,0 +1,569 @@
...
+	Concurrency            uint     `help:"Maximum number of packages to process concurrently. Setting it to 0 puts no limit on the concurrency, i.e., all packages are processed in parallel." default:"0"`
+	PushRetry              uint     `help:"Number of retries when pushing a provider package fails." default:"3"`
+
+	Platform        []string `help:"Platforms to build the packages for. Each platform should use the <OS>_<arch> syntax. An example is: linux_arm64." default:"linux_amd64,linux_arm64"`
+	ProviderBinRoot string   `short:"p" help:"Provider binary paths root. Smaller provider binaries should reside under the platform directories in this folder." type:"existingdir"`
+	Output          string   `short:"o" help:"Path of the package output directory." optional:""`
```
turkenh, line 105:
```suggestion
	OutputDir          string   `short:"o" help:"Path of the package output directory." optional:""`
```

nit: more explicit that this is a directory.

cmd/up/xpkg/batch.go
```diff
@@ -0,0 +1,569 @@
...
+	StorePackage    []string `help:"Service name whose provider package should be stored under the package output directory specified with the --output option." optional:""`
```
turkenh, line 106:
```suggestion
	StoreSubpackages    []string `help:"Package names whose providers should be stored under the package output directory specified with the --output-dir option." optional:""`
```
Also related to naming `Services` vs `Subpackages`, there seems to be an inconsistency in how we are naming these smaller provider groups.

cmd/up/xpkg/batch.go
ulucinar: I like the term `subpackage` as it has connotations of subsets and subgraphs, referring to the more theoretical work we have done on this subject (where we had previously examined various metrics & statistics regarding the subgraphs of the cross-resource reference graph). So moving from a more theoretical plane to a more practical one, it made sense for me to switch also the terminology, i.e., from `subgraphs` to `subpackages`. 

But we have also deprecated the subpackage terminology and that's why I felt the need to use something else. According to our current conventions, we may be using something like `--smaller-providers` or maybe just `--providers` but the concept of a provider may be a little bit ambiguous in my opinion. Are we referring to different provider families (e.g., `provider-aws` and `provider-gcp`) or different members within the same family (which is the correct context in this case). 

So, to be precise and to reduce ambiguity, I believe it makes sense to use `--smaller-providers`. This will also be consistent with the previous changes we made.

ulucinar: Done. We sync'ed offline with @turkenh, and decided on a consistent naming across the provided command-line options.

cmd/up/xpkg/xpkg.go
ulucinar: Done.

cmd/up/xpkg/batch.go
ulucinar: Done.

cmd/up/xpkg/batch.go
ulucinar: Done per our offline discussion.

PR #343, opened 2023-07-10
Prevent deferred File.Close and wrap errors while putting CRDs into xpls provider package cache

ulucinar:

Removes the deferred `File.Close` call in `cache.entry.writeObjects` to decrease the # of simultaneously open files and wraps the error returned from `cache.entry.appendToPackageJSON` to provide more context.

I have:

Tested with Go 1.18.10 on a test configuration package with the following command:
```shell
/tmp/up xpkg dep xpkg.upbound.io/upbound/provider-gcp
```
(I've initially observed this issue with `up@v.17.0`, which uses Go `1.18`)

### DIFF
--- a/internal/xpkg/dep/cache/entry.go
+++ b/internal/xpkg/dep/cache/entry.go
@@ -36,10 +36,11 @@ const (
 	crdNameFmt = "%s.yaml"
 	delim      = "\n"
 
-	errFailedToCreateMeta      = "failed to create meta file in entry"
-	errFailedToCreateImageMeta = "faile to create image meta entry"
-	errFailedToCreateCRD       = "failed to create crd"
-	errNoObjectsToFlushToDisk  = "no objects to flush"
+	errFailedToCreateMeta          = "failed to create meta file in entry"
+	errFailedToCreateImageMeta     = "failed to create image meta entry"
+	errFailedToCreateCacheEntryFmt = "failed to create a cache entry for the CRD at path %s"
+	errFailedToAppendCRDFmt        = "failed to add CRD %q to package"
+	errNoObjectsToFlushToDisk      = "no objects to flush"
 )
 
 // entry is the internal representation of the cache at a given directory
@@ -246,23 +247,13 @@ func (e *entry) writeObjects(objs []runtime.Object) (*flushstats, error) { // no
 			continue
 		}
 
-		f, err := e.fs.Create(filepath.Join(e.location(), fmt.Sprintf(crdNameFmt, name)))
-		if err != nil {
-			return stats, err
-		}
-		defer f.Close() // nolint:errcheck
-
-		fb, err := f.Write(yb)
-		if err != nil {
-			return stats, err
-		}
-
-		if fb == 0 {
-			return stats, errors.New(errFailedToCreateCRD)
+		entryLocation := filepath.Join(e.location(), fmt.Sprintf(crdNameFmt, name))
+		if err := afero.WriteFile(e.fs, entryLocation, yb, 0o600); err != nil {
+			return stats, errors.Wrapf(err, errFailedToCreateCacheEntryFmt, entryLocation)
 		}
 
 		if err := e.appendToPackageJSON(jb); err != nil {
-			return stats, errors.New(errFailedToCreateCRD)
+			return stats, errors.Wrapf(err, errFailedToAppendCRDFmt, name)
 		}
 
 		inc()


internal/xpkg/dep/cache/entry.go
```diff
@@ -246,23 +246,12 @@ func (e *entry) writeObjects(objs []runtime.Object) (*flushstats, error) { // no
...
-
-		fb, err := f.Write(yb)
-		if err != nil {
-			return stats, err
-		}
-
-		if fb == 0 {
-			return stats, errors.New(errFailedToCreateCRD)
+		if err := os.WriteFile(filepath.Join(e.location(), fmt.Sprintf(crdNameFmt, name)), yb, 0o600); err != nil {
+			return stats, errors.Wrap(err, errFailedToCreateCRD)
```
sttts, line 250: Have you double checked that the filename is part of the errors from `os.WriteFile`?

internal/xpkg/dep/cache/entry.go
```diff
@@ -246,23 +246,12 @@ func (e *entry) writeObjects(objs []runtime.Object) (*flushstats, error) { // no
...
 		}
 
 		if err := e.appendToPackageJSON(jb); err != nil {
-			return stats, errors.New(errFailedToCreateCRD)
+			return stats, errors.Wrap(err, errFailedToCreateCRD)
```
sttts, line 254: that error is `failed to create crd`. Can a user understand what it means?

internal/xpkg/dep/cache/entry.go
```diff
@@ -246,23 +246,12 @@ func (e *entry) writeObjects(objs []runtime.Object) (*flushstats, error) { // no
...
-		defer f.Close() // nolint:errcheck
-
-		fb, err := f.Write(yb)
-		if err != nil {
-			return stats, err
-		}
-
-		if fb == 0 {
-			return stats, errors.New(errFailedToCreateCRD)
+		if err := os.WriteFile(filepath.Join(e.location(), fmt.Sprintf(crdNameFmt, name)), yb, 0o600); err != nil {
```
tnthornton, line 249: I think `os.Writefile` is going to side step the use of `afero.Fs` here. We can fix that up by using afero's API xref https://github.com/spf13/afero#using-aferos-utility-functions. So something like this:

```suggestion
		if err := afero.WriteFile(e.fs, filepath.Join(e.location(), fmt.Sprintf(crdNameFmt, name)), yb, 0o600); err != nil {
```
may work. Here's another example of doing that in this repo https://github.com/upbound/up/blob/main/internal/xpkg/build_test.go#L171.

internal/xpkg/dep/cache/entry.go
```diff
@@ -36,10 +36,11 @@ const (
 	crdNameFmt = "%s.yaml"
 	delim      = "\n"
 
-	errFailedToCreateMeta      = "failed to create meta file in entry"
-	errFailedToCreateImageMeta = "faile to create image meta entry"
-	errFailedToCreateCRD       = "failed to create crd"
-	errNoObjectsToFlushToDisk  = "no objects to flush"
+	errFailedToCreateMeta       = "failed to create meta file in entry"
+	errFailedToCreateImageMeta  = "failed to create image meta entry"
+	errFailedToCreateCacheEntry = "failed to create a cache entry for the CRD at path: %s"
```
sttts, line 41: nit: I believe to have seen a `Fmt` suffix in most places if it is a format.

internal/xpkg/dep/cache/entry.go
```diff
@@ -36,10 +36,11 @@ const (
...
+	errFailedToAppendCRD        = "failed to add the CRD to the package: %s"
```
sttts, line 42: this won't look good with `errors.Wrapf`:  `failed to add the CRD to the package: package-name: some error`. Would rather use:
```suggestion
	errFailedToAppendCRD        = "failed to add the CRD to package %q"
```

internal/xpkg/dep/cache/entry.go
ulucinar: Hi @sttts,
I think the file being processed is not always part of the error message coming from `os.WriteFile`. In my case, for instance, it was just `too many open files`. Added the cache entry's path into the error's context.

internal/xpkg/dep/cache/entry.go
ulucinar: Done. I've replaced that error message as `failed to create a cache entry for the CRD at path: %s`. I did not include the CRD name in the error message as the filename of the cache entry already has that information.

internal/xpkg/dep/cache/entry.go
ulucinar: Hi @tnthornton,
Thanks, done.

internal/xpkg/dep/cache/entry.go
ulucinar: Done.

tnthornton (approved):
Thanks @ulucinar ! My concern was addressed, so LGTM 👍

PR #347, opened 2023-08-07
Add AWS support to `up spaces billing get`

nlinx:
- Adds AWS support to `up spaces billing get`

Fixes #
https://github.com/upbound/mxe/issues/143
I have:

- Unit tests for clientutil functions
- Manual testing against Upbound test bucket with usage data within it

### DIFF
--- a/cmd/up/space/billing/get.go
+++ b/cmd/up/space/billing/get.go
@@ -27,13 +27,8 @@ import (
 	"strings"
 	"time"
 
-	gcs "cloud.google.com/go/storage"
 	"github.com/alecthomas/kong"
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/session"
-	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/crossplane/crossplane-runtime/pkg/errors"
-	gcpopt "google.golang.org/api/option"
 
 	"github.com/upbound/up/internal/usage"
 	"github.com/upbound/up/internal/usage/report"
@@ -174,7 +169,7 @@ func (c *getCmd) cleanupOnError() {
 	}
 }
 
-func (c *getCmd) collectReport() error { //nolint:gocyclo
+func (c *getCmd) collectReport() error {
 	f, err := os.Create(c.outAbs)
 	if err != nil {
 		return errors.Wrap(err, "error creating report")
@@ -199,32 +194,11 @@ func (c *getCmd) collectReport() error { //nolint:gocyclo
 	// TODO(branden): Add support for Azure.
 	switch {
 	case c.Provider == providerGCP:
-		opts := []gcpopt.ClientOption{}
-		if c.Endpoint != "" {
-			opts = append(opts, gcpopt.WithEndpoint(c.Endpoint))
-		}
-		gcsCli, err := gcs.NewClient(ctx, opts...)
-		if err != nil {
-			return errors.Wrap(err, "error creating storage client")
-		}
-		bkt := gcsCli.Bucket(c.Bucket)
-		if err := reportgcs.MaxResourceCountPerGVKPerMCP(ctx, c.Account, bkt, c.billingPeriod, time.Hour, rw); err != nil {
+		if err := reportgcs.GenerateReport(ctx, c.Account, c.Endpoint, c.Bucket, c.billingPeriod, time.Hour, rw); err != nil {
 			return err
 		}
 	case c.Provider == providerAWS:
-		sess, err := session.NewSession(&aws.Config{})
-		if err != nil {
-			return errors.Wrap(err, "error creating aws session")
-		}
-		config := &aws.Config{}
-		if c.Endpoint != "" {
-			config = &aws.Config{
-				Endpoint: aws.String(c.Endpoint),
-			}
-		}
-		s3client := s3.New(sess, config)
-
-		if err := reportaws.MaxResourceCountPerGVKPerMCP(ctx, c.Account, c.Bucket, s3client, c.billingPeriod, rw); err != nil {
+		if err := reportaws.GenerateReport(ctx, c.Account, c.Endpoint, c.Bucket, c.billingPeriod, rw); err != nil {
 			return err
 		}
 	default:
--- a/cmd/up/space/billing/get.go
+++ b/cmd/up/space/billing/get.go
@@ -29,11 +29,15 @@ import (
 
 	gcs "cloud.google.com/go/storage"
 	"github.com/alecthomas/kong"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/crossplane/crossplane-runtime/pkg/errors"
 	gcpopt "google.golang.org/api/option"
 
 	"github.com/upbound/up/internal/usage"
 	"github.com/upbound/up/internal/usage/report"
+	reportaws "github.com/upbound/up/internal/usage/report/aws"
 	reporttar "github.com/upbound/up/internal/usage/report/file/tar"
 	reportgcs "github.com/upbound/up/internal/usage/report/gcs"
 )
@@ -78,9 +82,11 @@ func (d *dateRange) Decode(ctx *kong.DecodeContext) error {
 type provider string
 
 func (p provider) Validate() error {
-	// TODO(branden): Add support for AWS and Azure.
+	// TODO(branden): Add support Azure.
 	switch p {
 	case providerGCP:
+		return nil
+	case providerAWS:
 		return nil
 	default:
 		return fmt.Errorf(errFmtProviderNotSupported, p)
@@ -168,7 +174,7 @@ func (c *getCmd) cleanupOnError() {
 	}
 }
 
-func (c *getCmd) collectReport() error {
+func (c *getCmd) collectReport() error { //nolint:gocyclo
 	f, err := os.Create(c.outAbs)
 	if err != nil {
 		return errors.Wrap(err, "error creating report")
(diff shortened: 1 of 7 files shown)


cmd/up/space/billing/get.go
```diff
@@ -82,6 +86,8 @@ func (p provider) Validate() error {
 	switch p {
 	case providerGCP:
 		return nil
+	case providerAWS:
+		return nil
```
AlainRoy, line 90: If adding support for AWS, it'd be appropriate to update the // TODO on line 85 that says we need to add support for AWS.

nlinx: updated

internal/usage/report/aws/aws.go
```diff
@@ -0,0 +1,118 @@
...
+			if err != nil {
+				break
+			}
+			g.Go(func() error {
+				return readObject(ctx, ag, agMu, resp)
+			})
+		}
+		if err := g.Wait(); err != nil {
+			return errors.Wrap(err, errReadEvents)
+		}
```
branden, line 83: It looks like this code intends to download objects in parallel, but I don't believe that's what's happening. From [docs](https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.44.318/service/s3#S3.GetObject) it looks like `client.GetObject()` is blocking and downloading the entire object before it returns. Since that's happening in the main goroutine, that means we're downloading objects in serial. The worker goroutines are only decoding the already downloaded objects, which probably isn't helping us to increase network utilization.

I see a couple ways forward:

1. Rewrite this section to use [s3manager](https://pkg.go.dev/github.com/aws/aws-sdk-go/service/s3@v1.44.318#hdr-Download_Manager), which offers a concurrent download manager.
2. Stick with downloading objects in serial and remove the unnecessary concurrency. This could mean slow downloads, but we can optimize that later. If we go this route, let's make sure to use [GetObjectWithContext](https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.44.318/service/s3#S3.GetObjectWithContext) so that downloads can be cancelled if e.g. the user ctrl+c's the command.

internal/usage/report/aws/aws.go
```diff
@@ -0,0 +1,118 @@
...
+			if err := w.Write(e); err != nil {
+				return errors.Wrap(err, errWriteEvents)
+			}
+		}
+	}
+	return nil
+}
+
+// readObject() decodes MCP GVK events from an object and adds them to an aggregate.
+func readObject(ctx context.Context, ag *aggregate.MaxResourceCountPerGVKPerMCP, agMu sync.Locker, obj *s3.GetObjectOutput) error { // nolint:unparam
```
branden, line 97: If we get rid of the unused `ctx` arg here then we can also remove the `nolint`.

branden (commented):
Thanks @nlinx! I left some comments, I think there might be bugs regarding storage queries and download concurrency. But they look quite fixable, this looks close.

cmd/up/space/billing/get.go
```diff
@@ -191,7 +197,8 @@ func (c *getCmd) collectReport() error {
 	defer stop()
 
 	// TODO(branden): Add support for AWS and Azure.
```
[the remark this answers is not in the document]
branden: Oops just saw this is already resolved!

internal/usage/report/aws/aws.go
nlinx: Ooo good catch! Wouldn't we achieve the async attributes we want by moving the GetObject call into the go routine (line 77)?

internal/usage/report/aws/aws.go
```diff
@@ -0,0 +1,118 @@
...
+
+	for iter.More() {
+		startPrefix, _, start, end, err := iter.Next()
+		if err != nil {
+			return errors.Wrap(err, errReadEvents)
+		}
+		objects, err := client.ListObjectsV2(&s3.ListObjectsV2Input{
+			Bucket: aws.String(bucket),
+			Prefix: aws.String(startPrefix),
+		})
```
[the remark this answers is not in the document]
nlinx: The startPrefix returns as a string in this format `"account=%s/date=%s/hour=%02d/"`. The prefix is the base of what's being fetched. i.e. it'll fetch everything where the path starts with that. Since it includes the hour in there an "end" isn't necessary for our use case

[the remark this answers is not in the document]
branden: Our use case (currently) only needs 1h windows, but as it stands this function has a bug based on its docstring and inputs, which offer a flexible time window. I'd prefer that we fix the bug, but updating the function to make clear that it only supports 1h windows would help.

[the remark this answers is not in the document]
nlinx: I think there may be a misunderstanding here. This function should still go through everything in our time range based on the iterator.

The ListObjectsV2 call is within the iterator loop so we'll ultimately be calling ListObjectV2 on every `account=%s/date=%s/hour=%02d/` combo the iterator returns to us.

internal/usage/report/aws/aws.go
nlinx: Ok I've decided the right path is to use GetObjectWithContext and move the call into the go routine. My primary reason is so we can reuse existing code. Utilizing s3manager requires me to write to a file; and while that is ultimately what we're doing, we already have code that is going to be shared in usage with GCS support. Reusing our existing code allows us to avoid having a unique write path for AWS

branden: I'm fine with `GetObjectWithContext()` if it works. FWIW you don't need to download to a file to use s3manager.

https://docs.aws.amazon.com/sdk-for-go/api/service/s3/s3manager/#Downloader.DownloadWithContext

> The w io.WriterAt can be satisfied by an os.File to do multipart concurrent downloads, or in memory []byte wrapper using aws.WriteAtBuffer.

internal/usage/report/aws/aws.go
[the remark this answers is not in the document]
branden: That logic means we're skipping data if the window is longer than 1h. See this query iterator test case as an example: https://github.com/upbound/up/pull/347/files#diff-ea0f37c9233e73fdfadd963f178c84a17cc7a966cc9d31eade20ab7b4605d302R188-R210

```
		"3HourRange2HourWindow": {
			reason: "3h range divided into 2h windows.",
			args: args{
				account: "test-account",
				start:   time.Date(2006, 5, 4, 3, 0, 0, 0, time.UTC),
				end:     time.Date(2006, 5, 4, 6, 0, 0, 0, time.UTC),
				window:  2 * time.Hour,
			},
			want: []iteration{
				{
					StartOffset: "account=test-account/date=2006-05-04/hour=03/",
					EndOffset:   "account=test-account/date=2006-05-04/hour=05/",
					Start:       time.Date(2006, 5, 4, 3, 0, 0, 0, time.UTC),
					End:         time.Date(2006, 5, 4, 5, 0, 0, 0, time.UTC),
				},
				{
					StartOffset: "account=test-account/date=2006-05-04/hour=05/",
					EndOffset:   "account=test-account/date=2006-05-04/hour=06/",
					Start:       time.Date(2006, 5, 4, 5, 0, 0, 0, time.UTC),
					End:         time.Date(2006, 5, 4, 6, 0, 0, 0, time.UTC),
				},
			},
		},
```

This test case is grouping a 3-hour range (covering hours `03`, `04`, and `05`) into 2-hour windows. Note that there is no `StartOffset` containing `hour=04`. If we only looked at data prefixed with the `StartOffset` values, we'd be skipping the data for that hour.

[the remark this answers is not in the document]
nlinx: ahhh ok the misunderstanding was on my end since I was only thinking about the 1hr window case. you're right. we we use a larger window than an hour the logic breaks down. i'll add a todo and update the function description to note that we will only support one hour windows.

cmd/up/space/billing/get.go
```diff
@@ -168,7 +174,7 @@ func (c *getCmd) cleanupOnError() {
 	}
 }
 
-func (c *getCmd) collectReport() error {
+func (c *getCmd) collectReport() error { //nolint:gocyclo
```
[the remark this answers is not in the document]
branden: Bump, let's refactor this method instead of silencing the gocyclo linter on it.

branden (commented):
This LGTM, there's just one lint error that I think we should address rather than silencing it.

internal/usage/report/gcs/gcs.go
```diff
@@ -39,10 +41,27 @@ const (
 	errWriteEvents = "error writing events"
 )
 
-// MaxResourceCountPerGVKPerMCP reads usage data for an account and time range
+// GenerateReport
```
branden, line 44: nit: Incomplete docstring

internal/usage/report/aws/aws.go
```diff
@@ -40,10 +41,29 @@ const (
 	errWriteEvents = "error writing events"
 )
 
-// MaxResourceCountPerGVKPerMCP reads usage data for an account and time range
+func GenerateReport(ctx context.Context, account, endpoint, bucket string, billingPeriod usage.TimeRange, w report.MCPGVKEventWriter) error {
```
branden, line 44: nit: No docstring

internal/usage/report/gcs/gcs.go
```diff
@@ -20,9 +20,11 @@ import (
 	"time"
 
 	"cloud.google.com/go/storage"
+
```
branden, line 23: nit: `cloud.google.com/go/storage` should be grouped with the other third-party imports below.

branden (approved):

internal/usage/report/gcs/gcs.go
nlinx: lol that's what i get for stepping away to eat while writing doc string

branden (approved):

PR #351, opened 2023-08-11
Add Azure support to `up space billing get` + provider-generic refactor

branden:

This adds support for Azure blob storage to `up space billing get`. It also refactors the GCS and S3 implementations to use provider-generic code for generating the report.

Other things worth noting:
* All provider implementations support reading gzipped data.
* S3 events can now be grouped by arbitrarily sized windows, like the other providers.
* The number of S3 objects that can be read in a time window is now unlimited, up from 5000.
* I gave no consideration to optimizing download speed with concurrency. I'll leave this for a later PR.

This is a big PR, so see commits for logical groupings of changes.

I have:

I ran `up space billing get` against sample data with each storage provider. I'll update this with more detail shortly.


internal/usage/event/reader/reader.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"context"
+	"errors"
+
+	"github.com/upbound/up/internal/usage/event"
+	"github.com/upbound/up/internal/usage/model"
+)
+
+var EOF = event.EOF
```
darkmuggle, line 25: Seems like this error should be a common error?

darkmuggle (approved):
Minior nits, but otherwise LGTM

internal/usage/event/reader/reader.go
branden: I'm not sure what you mean, can you elaborate?

PR #360, opened 2023-08-15
Spaces: add support for public load balancer

darkmuggle:

* Improve detection of Kubernetes providers for Spaces
* Provision public LB on managed Kubernetes

I have:

### DIFF
--- a/cmd/up/space/defaults/defaults.go
+++ b/cmd/up/space/defaults/defaults.go
@@ -16,7 +16,7 @@ package defaults
 
 import (
 	"context"
-	"fmt"
+	"errors"
 	"strings"
 
 	"github.com/pterm/pterm"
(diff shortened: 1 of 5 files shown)


cmd/up/space/defaults/defaults.go
```diff
@@ -15,74 +15,104 @@
...
+	return &CloudConfig{
+		SpacesValues:  cloud.Defaults().SpacesValues,
+		PublicIngress: cloud.Defaults().PublicIngress,
+	}, nil
+}
+
+func detectKubernetes(kClient kubernetes.Interface) (CloudType, error) {
+	// EKS and Kind are _harder_ to detect based on version, so look at node labels.
+	ctx := context.Background()
+	if nodes, err := kClient.CoreV1().Nodes().List(ctx, metav1.ListOptions{}); err == nil {
```
tnthornton, line 101: Do we want to return the error if it's non-nil?

cmd/up/space/init.go
```diff
@@ -182,8 +207,9 @@ type initCmd struct {
 	token      string
 	quiet      config.QuietFlag
 
-	Version string `arg:"" help:"Upbound Spaces version to install."`
-	Yes     bool   `name:"yes" type:"bool" help:"Answer yes to all questions"`
+	Version        string `arg:"" help:"Upbound Spaces version to install."`
+	Yes            bool   `name:"yes" type:"bool" help:"Answer yes to all questions"`
+	PrivateIngress bool   `name:"private-ingress" type:"bool" help:"For AKS,EKS,GKE do not expose ingress publically"`
```
tnthornton, line 212: I think this will default to false, do we want to default it to true?

cmd/up/space/init.go
```diff
@@ -115,19 +115,44 @@ func (c *initCmd) AfterApply(insCtx *install.Context, kongCtx *kong.Context, qui
...
+				skip = true
+			}
+		}
+		if !skip {
+			c.Set[k] = v
+		}
+	}
+	if c.PrivateIngress {
+		defs.PublicIngress = false
+	}
```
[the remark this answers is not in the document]
darkmuggle: Updated

cmd/up/space/defaults/defaults.go
```diff
@@ -15,74 +15,104 @@
...
-	if err != nil {
-		return s, err
-	}
-	if cloud == Generic {
-		pterm.Info.Println("Setting defaults for generic Kubernetes")
-		return s, nil
-	}
+var vendorDefaults = map[CloudType]map[string]string{
+	AmazonEKS: {
+		"clusterType": string(AmazonEKS),
```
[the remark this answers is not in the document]
darkmuggle: Fixed.

cmd/up/space/defaults/defaults.go
darkmuggle: Maybe? I didn't error since nodes is not a required aspect of Spaces. If the detection fails, I didn't think that this worthy to abort the installation since the cloud type could be set via `--set clusterType=`

tnthornton: OK. If we don't think it's worthy, maybe we could add a comment for why we don't care and then handle the lint err 👍

cmd/up/space/init.go
darkmuggle: I'd argue from a customer perspective, if you are in a public cloud that the ingress should be public by default (otherwise you have manually change that). The cloud logic sets it to false for Generic/undetected and Kind.

darkmuggle:
Tested and confirmed to work with GKE, AKS, EKS and kind -- all work beautifully

cmd/up/space/init.go
tnthornton: The few customers I've talked to have generally been surprised when we have defaulted to setting up a public loadbalancer. Maybe the sampleset is skewed, but it seems like we should default to the common case and allow for opting out in the uncommon case (allowing a public loadbalancer).

darkmuggle: Fair enough....the last push is now:

tnthornton (approved):
One last small nit, otherwsie LGTM.

PR #362, opened 2023-08-18
Space: improve destroy command

darkmuggle:

For Spaces destroy command:
- Remove the `upbound-system` namespace on destroy
- Make the user confirm the delete

I have:

Tested against v0.15.1 with Spaces

```

******************** DESTRUCTIVE COMMAND ********************
****************** DATA-DESTRUCTION WARNING *****************

 WARNING  Destroying Spaces is a destructive command that will destroy data and oprhan resources.
 WARNING  Before proceeding ensure that Managed Resources in Control Planes have been deleted.
 WARNING  All Spaces components including Control Planes will be destroyed.

To proceed, type: "DESTROY CONFIRMED": DESTROY CONFIRMED
```

### DIFF
--- a/cmd/up/space/destroy.go
+++ b/cmd/up/space/destroy.go
@@ -15,18 +15,45 @@
 package space
 
 import (
+	"fmt"
+	"os"
+
 	"github.com/pterm/pterm"
 
+	"github.com/upbound/up/internal/input"
 	"github.com/upbound/up/internal/install"
 	"github.com/upbound/up/internal/install/helm"
 	"github.com/upbound/up/internal/upterm"
 )
+
+const confirmStr = "DESTROY CONFIRMED"
 
 // AfterApply sets default values in command after assignment and validation.
 func (c *destroyCmd) AfterApply(insCtx *install.Context) error {
 	// NOTE(tnthornton) we currently only have support for stylized output.
 	pterm.EnableStyling()
 	upterm.DefaultObjPrinter.Pretty = true
+
+	if !c.Confirmed {
+		prompter := input.NewPrompter()
+		pterm.Println()
+		pterm.FgRed.Println("******************** DESTRUCTIVE COMMAND ********************")
+		pterm.FgRed.Println("****************** DATA-DESTRUCTION WARNING *****************")
+		pterm.Println()
+		pterm.Warning.Println("Destroying Spaces is a destructive command that will destroy data and oprhan resources.")
+		pterm.Warning.Println("Before proceeding ensure that Managed Resources in Control Planes have been deleted.")
+		pterm.Warning.Println("All Spaces components including Control Planes will be destroyed.")
+		pterm.Println()
+		in, err := prompter.Prompt(fmt.Sprintf("To proceed, type: %q", confirmStr), false)
+		if err != nil {
+			pterm.Error.Printfln("error getting user confirmation: %v", err)
+			os.Exit(1)
+		}
+		if in != confirmStr {
+			pterm.Error.Println("Destruction was not confirmed")
+			os.Exit(10)
+		}
+	}
 
 	mgr, err := helm.NewManager(insCtx.Kubeconfig,
 		spacesChart,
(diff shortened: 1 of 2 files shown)


cmd/up/space/destroy.go
```diff
@@ -15,19 +15,58 @@
...
 func (c *destroyCmd) AfterApply(insCtx *install.Context) error {
 	// NOTE(tnthornton) we currently only have support for stylized output.
 	pterm.EnableStyling()
 	upterm.DefaultObjPrinter.Pretty = true
 
+	if !c.Confirmed {
+		prompter := input.NewPrompter()
+		pterm.Println()
+		pterm.FgRed.Println("******************** DESTRUCTIVE COMMAND ********************")
+		pterm.FgRed.Println("****************** DATA-DESTRUCTION WARNING *****************")
```
AlainRoy, line 47: Nit: Maybe "DATA LOSS WARNING" instead of destruction. Only because it sounds a bit less repetitive with the first command.

I know, that's the tiniest nit ever. Feel free to ignore me, it's Friday afternoon and this is what I can muster.

cmd/up/space/destroy.go
```diff
@@ -42,12 +81,18 @@ func (c *destroyCmd) AfterApply(insCtx *install.Context) error {
 
 // destroyCmd uninstalls Upbound.
 type destroyCmd struct {
-	mgr install.Manager
+	mgr     install.Manager
+	kClient kubernetes.Interface
 
 	commonParams
+
+	Confirmed bool `name:"yes-really-delete-spaces-and-all-data" type:"bool" help:"Bypass safety checks and destroy Spaces"`
```
AlainRoy, line 89: I love it!

AlainRoy: Well, I love it, but it probably should be singular -- space instead of spaces.

AlainRoy (approved):
I left nits, but it looks good to me.

tnthornton:
Do we want to pipe in allowing them to orphan their resources?

cmd/up/space/destroy.go
AaronME: Users might also accept `force` :-)

tnthornton: I personally like `force` since it follows other conventions 🤷

cmd/up/space/destroy.go
darkmuggle: Pfft, I love the bike shed. And yeah, it does studder a bit, I'll fix it.

cmd/up/space/destroy.go
```diff
@@ -42,12 +81,18 @@ func (c *destroyCmd) AfterApply(insCtx *install.Context) error {
...
 }
 
 // Run executes the uninstall command.
 func (c *destroyCmd) Run(insCtx *install.Context) error {
-	return c.mgr.Uninstall()
+	if err := c.mgr.Uninstall(); err != nil {
+		return err
+	}
+	return c.kClient.CoreV1().Namespaces().Delete(context.Background(), nsUpboundSystem, v1.DeleteOptions{})
```
branden, line 97: This would delete the crossplane and upbound-pull-secret that we ask users to create before deploying. Is that OK? I could at least see it making testing more tedious. Maybe we could introduce a flag to skip deleting the namespace?

cmd/up/space/destroy.go
darkmuggle: Well I _like_ `force` and it does follow other conventions, I wanted to make people think.

cmd/up/space/destroy.go
darkmuggle: Check the latest push -- we only delete the namespace if there are no orphans.

darkmuggle:
> Do we want to pipe in allowing them to orphan their resources?

Yeah, implemented in the last push. The way I did it was to run not execute any hooks when orphan is set. The other way would be to upgrade and set the deletionPolicy, however that doesn't work with out a token file; it seemed crazy to require a token file to _remove_ something. Since the delete is part of a hook, disabling the hook accomplished the same goal. 

If you `up space destroy --orphan` and then `up space init` things start working again.

darkmuggle:
@tnthornton @branden if the last drop looks good, go ahead and merge.

PR #376, opened 2023-09-26
spaces: add bidirectional version constraints

sttts:
1. `cmd/up/space/version.go`: static semver constraints against the chart
2. `spaces.upbound.io/up-version-constraints` annotation on the chart: constraints against `up`.

Fixes https://github.com/upbound/mxe/issues/190.

### DIFF
--- a/cmd/up/space/init.go
+++ b/cmd/up/space/init.go
@@ -27,6 +27,7 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
 	"github.com/pterm/pterm"
 	"golang.org/x/exp/maps"
+	"helm.sh/helm/v3/pkg/chart"
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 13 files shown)


branden (approved):
LGTM. It looks like since we aren't yet enforcing a less-than version constraint in `up`, we won't yet need to cut a new release with updated constraints when we release Spaces.

sttts:
>  It looks like since we aren't yet enforcing a less-than version constraint in up, we won't yet need to cut a new release with updated constraints when we release Spaces.

Right.

But I added https://github.com/upbound/up/pull/376/files#diff-73c5a5c16c72949218b0a84e291b541d8cf3305dc046be9b6ee06005bcb89ba1R46, but merely as an example. Not sure we have a reason yet. @tnthornton might know more.

sttts:
Updated to not require v1.1 for now.

AlainRoy:
@sttts  -- This looks good, thanks!

To clarify though, I think we also need to make annotations on the charts. Is that right? If so, what do we need to do to make sure that we keep those annotations up-to-date when we do a release?

PR #382, opened 2023-10-05
fix auth annotation not being enabled on v1

nullable-eth:
Providers moving from v1alpha1 do not get annotated with the auth configuration.  This change accounts for v1alpha1 and v1 providers annotating the crd with the upbound console config.

Fixes #1298

I have:

Tested locally building v1 and v1alpha1 provider with auth extension file, package was annotated correctly in both cases.

### DIFF
--- a/internal/xpkg/build.go
+++ b/internal/xpkg/build.go
@@ -197,39 +197,48 @@ func (b *Builder) Build(ctx context.Context, opts ...BuildOpt) (v1.Image, runtim
 		linter = NewFunctionLinter()
 	case pkgmetav1.ProviderKind:
 		if b.ab != nil { // if we have an auth.yaml file
+			authGroup := ""
 			if p, ok := meta.(*v1alpha1.Provider); ok {
-				// if has annotation auth.upbound.io/group then look for the object
+				if group, ok := p.ObjectMeta.Annotations[authMetaAnno]; ok {
+					authGroup = group
+				}
+			}
+			if p, ok := meta.(*pkgmetav1.Provider); ok {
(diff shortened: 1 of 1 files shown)


internal/xpkg/build.go
```diff
@@ -197,39 +197,48 @@ func (b *Builder) Build(ctx context.Context, opts ...BuildOpt) (v1.Image, runtim
 		linter = NewFunctionLinter()
 	case pkgmetav1.ProviderKind:
 		if b.ab != nil { // if we have an auth.yaml file
+			authGroup := ""
 			if p, ok := meta.(*v1alpha1.Provider); ok {
```
tnthornton, line 201:
```suggestion
			if p, ok := meta.(metav1.Object); ok {
```

internal/xpkg/build.go
```diff
@@ -197,39 +197,48 @@ func (b *Builder) Build(ctx context.Context, opts ...BuildOpt) (v1.Image, runtim
...
-				// if has annotation auth.upbound.io/group then look for the object
-				// specified there like aws.upbound.io and annotate that with auth.upbound.io/config
-				// and embed the contents of the auth.yaml file
 				if group, ok := p.ObjectMeta.Annotations[authMetaAnno]; ok {
```
tnthornton, line 202:
```suggestion
				if group, ok := p.GetAnnotations()[authMetaAnno]; ok {
```

tnthornton (commented):
Thanks @nullable-eth. In my local tests, updating using the suggestions I have below allows you to have just the 1 cast instead of introducing one for each type.

This works for now since we're only doing a type cast to get to the annotations 👍.

tnthornton (approved):

PR #384, opened 2023-10-05
Add tests for auth.yaml inclusion in provider package

tnthornton:
This is a follow on to https://github.com/upbound/up/pull/382. Prior to this change there wasn't test coverage over this specific path.

I have:

#### Before including #382 
make test - verified that the `SuccessAuthV1Provider` test failed.

#### After including #382 
make test - verified tests passed


internal/xpkg/build_test.go
```diff
@@ -305,6 +456,15 @@ func readImg(i v1.Image) (xpkgContents, error) {
 		return contents, err
 	}
 	contents.pkgBytes = pkgBytes
+	ps := string(pkgBytes)
+
+	// This is pretty unfortunate. Unless we build out steps to re-parse the
+	// package from the image (i.e. the system under test) we're left
+	// performing string parsing. For now we choose part of the auth spec,
+	// specifically the version and date used in auth yamls.
```
AlainRoy, line 464: Instead of "auth yamls" (which sounds generic), do you mean "testdata/auth.yaml"? I know it's a nit, just suggesting a clarification if someone is reading it and trying to figure out which yamls you're referring to.

AlainRoy (approved):
Nice work!

internal/xpkg/build_test.go
tnthornton: No I was referring to the generic term. We refer to them internally as auth yaml's; let me see if there's a different name that was used in the design doc.

AlainRoy: OK, my bad -- disregard my comment.

tnthornton: No worries! I just looked at the design doc and there isn't a descriptive term used specifically for this file 😞.

PR #386, opened 2023-10-06
add: TarPath parameter to the build command

### LINKED ISSUE #385: Add support for local .tar images in xpkg build
We're using `podman` in our pipelines, and unfortunately local built tar images isn't supported (as seen [here](https://github.com/upbound/up/blob/main/cmd/up/xpkg/build.go#L95-L97)). 

Using your function [`xpkgFetch`](https://github.com/upbound/up/blob/main/cmd/up/xpkg/xpextract.go#L72-L76), and a hard-coded path as seen [here](https://github.com/upbound/up/commit/9243c6a8ceb00869c6ec355fe95782b45786abcc) we were able to build using `podman`. The flow currently looks like this:

```
podman build .
      --file ci/docker/Dockerfile
      --tag $IMAGE_NAME:latest
podman save -o local_image.tar $IMAGE_NAME:latest
./up xpkg build
      --package-root package
      --controller $IMAGE_NAME:latest
      --output my_provider.xpkg
```

Add a parameter that allow us to specify a local `.tar`.

JonasKs:

Add a `--tar-path` parameter, which is an alternative to `--controller`. This allow us to use `podman` in our pipelines. Example working flow:

```
podman build .
      --file ci/docker/Dockerfile
      --tag $IMAGE_NAME:latest
podman save -o local_image.tar $IMAGE_NAME:latest
./up xpkg build
      --package-root package
      --tar-path $IMAGE_NAME:latest
      --output my_provider.xpkg
```

Fixes #385 

I have:

We have tested a locally built binary, uploaded to a private s3 bucket in a GitLab CI pipeline, using `podman`, and it works as expected.

### DIFF
--- a/cmd/up/xpkg/build.go
+++ b/cmd/up/xpkg/build.go
@@ -92,10 +92,11 @@ func (c *buildCmd) AfterApply() error {
 		examples.New(),
 	)
 
-	// NOTE(hasheddan): we currently only support fetching controller image from
-	// daemon, but may opt to support additional sources in the future.
-	c.fetch = daemonFetch
-
+	if c.TarPath != "" {
+		c.fetch = xpkgFetch(c.TarPath)
+	} else {
+		c.fetch = daemonFetch
+	}
 	return nil
 }
 
@@ -113,6 +114,7 @@ type buildCmd struct {
 	ExamplesRoot string   `short:"e" help:"Path to package examples directory." default:"./examples"`
 	AuthExt      string   `short:"a" help:"Path to an authentication extension file." default:"auth.yaml"`
 	Ignore       []string `help:"Paths, specified relative to --package-root, to exclude from the package."`
+	TarPath      string   `help:"Path to tar file, an alternative to Controller."`
 }
 
 func (c *buildCmd) Help() string {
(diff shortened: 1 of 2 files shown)


cmd/up/xpkg/build.go
```diff
@@ -92,10 +92,11 @@ func (c *buildCmd) AfterApply() error {
...
 
-	// NOTE(hasheddan): we currently only support fetching controller image from
-	// daemon, but may opt to support additional sources in the future.
-	c.fetch = daemonFetch
-
+	if c.TarPath != "" {
+		c.fetch = xpkgFetch(c.TarPath)
+	} else {
+		c.fetch = daemonFetch
+	}
```
sttts, line 99: this needs a check that not both flags are used, doesn't it?

Wondering, xpkg is moving (or has moved) to crossplane's crank tool right now. We will need the same patch over there.

JonasKs: This has been refactored and merged in crossplane. Not sure if you'd like me to just close this one?

sttts: I guess medium term we will either drop this command here or vendor in crossplane's CLI tool. So I would suggest to at lesat open the same PR in parallel for crossplane such that we don't diverge.

sttts: Compare https://github.com/crossplane/crossplane/tree/master/cmd/crank/xpkg

PR #393, opened 2023-10-11
Re-introduce `up ctp connect` add `up ctp disconnect`

### LINKED ISSUE #391: Add `up ctp connect` and `up ctp disconnect` subcommands for cloud and spaces
Now that there has been some focus on bringing CLI functionality from Cloud on par with Spaces, a noticeable missing feature in both places in a nice, succinct, command that would allow an operator to easily "connect" or "disconnect" to/from a target control plane.

At a very high level this could be described as:
* `up ctp connect ctp-one` - this would acquire the kubeconfig for ctp-one, add it to the current kubeconfig and change the context to point to that control plane. That way any subsequent kubectl commands would be invoked against the control plane.
* `up ctp disconnect` - this would change the kube context back to the original context and any subsequent kubectl commands would be invoked against the original cluster.

tnthornton:
In order to make it easier to interact with a control plane, we're introducing a new command and bringing back a previous command with different behavior.

#### Connect
`up ctp connect {name}` has been reworked to connect the operators local kubeconfig to the target control plane. From a UX perspective, this means that an operator can simply point to a control plane and easily begin working within it.

#### Disconnect
`up ctp disconnect` has been introduced to easily allow the operator to switch back to the previous context they were using before initiating the `connect` command. This makes switching between target a control plane and another kubernetes cluster, quick and frictionless.

Fixes #391 

I have:

1. Introduced new tests
2. Captured the below video working against a local space
https://github.com/upbound/up/assets/2375126/1fbe5e6c-09e6-4289-9d75-a7f8a178aae6

3. Verified functionality was consistent between spaces and cloud.

### DIFF
--- a/cmd/up/controlplane/connect.go
+++ b/cmd/up/controlplane/connect.go
@@ -1,9 +1,190 @@
+// Copyright 2023 Upbound Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package controlplane
 
-// upbound_acct1_ctp1_kind-kind
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
 
-// use the current context in kubeconfig
-// config, err := clientcmd.BuildConfigFromFlags("", *kubeconfig)
-// if err != nil {
-// 	panic(err.Error())
-// }
+	"github.com/alecthomas/kong"
+	"github.com/pterm/pterm"
+
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/clientcmd/api"
+
+	"github.com/upbound/up-sdk-go/service/configurations"
+	cp "github.com/upbound/up-sdk-go/service/controlplanes"
+
+	"github.com/upbound/up/internal/controlplane"
+	"github.com/upbound/up/internal/controlplane/cloud"
+	"github.com/upbound/up/internal/controlplane/space"
+	"github.com/upbound/up/internal/kube"
+	"github.com/upbound/up/internal/upbound"
+	"github.com/upbound/up/internal/upterm"
+)
+
+var (
+	upboundPrefix = "upbound_"
+
+	errFmtConfigBroken = "config is broken, missing %s: %q"
+)
+
+type ctpConnector interface {
+	GetKubeConfig(ctx context.Context, name string) (*api.Config, error)
+}
+
+// AfterApply sets default values in command after assignment and validation.
+func (c *connectCmd) AfterApply(kongCtx *kong.Context, upCtx *upbound.Context) error {
+
+	if upCtx.Profile.IsSpace() {
+		kubeconfig, err := upCtx.Profile.GetKubeClientConfig()
+		if err != nil {
+			return err
+		}
+		client, err := dynamic.NewForConfig(kubeconfig)
+		if err != nil {
+			return err
+		}
+		c.client = space.New(client)
+	} else {
+		if c.Token == "" {
+			return fmt.Errorf("--token must be specified")
+		}
+
+		cfg, err := upCtx.BuildSDKConfig()
+		if err != nil {
+			return err
+		}
+		ctpclient := cp.NewClient(cfg)
+		cfgclient := configurations.NewClient(cfg)
+
+		// The cloud client needs the proxy endpoint and a PAT token for
+		// setting up communication with Upbound Cloud.
+		c.client = cloud.New(
+			ctpclient,
+			cfgclient,
+			upCtx.Account,
+			cloud.WithToken(c.Token),
+			cloud.WithProxyEndpoint(upCtx.ProxyEndpoint),
+		)
+	}
+
+	kongCtx.Bind(pterm.DefaultTable.WithWriter(kongCtx.Stdout).WithSeparator("   "))
+	return nil
+}
+
+// getCmd gets a single control plane in an account on Upbound.
+type connectCmd struct {
+	Name  string `arg:"" required:"" help:"Name of control plane." predictor:"ctps"`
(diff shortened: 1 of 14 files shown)


cmd/up/controlplane/connect.go
```diff
@@ -0,0 +1,202 @@
...
+
+	"github.com/upbound/up/internal/controlplane"
+	"github.com/upbound/up/internal/controlplane/cloud"
+	"github.com/upbound/up/internal/controlplane/space"
+	"github.com/upbound/up/internal/kube"
+	"github.com/upbound/up/internal/upbound"
+	"github.com/upbound/up/internal/upterm"
+)
+
+var (
```
darkmuggle, line 43:
```suggestion
const (
```

cmd/up/controlplane/disconnect.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/alecthomas/kong"
+	"github.com/pterm/pterm"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/clientcmd/api"
+
+	"github.com/upbound/up/internal/upbound"
+	"github.com/upbound/up/internal/upterm"
+)
+
+var (
```
darkmuggle, line 31:
```suggestion
const (
```

cmd/up/controlplane/connect.go
```diff
@@ -0,0 +1,202 @@
...
+			return err
+		}
+		client, err := dynamic.NewForConfig(kubeconfig)
+		if err != nil {
+			return err
+		}
+		c.client = space.New(client)
+	} else {
+		if c.Token == "" {
+			return fmt.Errorf("--token must be specified")
```
AlainRoy, line 69: This isn't a comment on the current PR, because I know the requirement for a token isn't new. But this requirement has always bothered me, from a UX perspective. I wonder if (one day, not now, probably not soon), we could find a way to not requirement a token. Maybe auto-create it? Maye have a class of tokens that expire, so they don't accumulate forever? 

I'm just thinking out loud--it's not a comment on this PR.

tnthornton: Agree 💯. I'm finding the UX of specifying this cumbersome.

cmd/up/space/init.go
```diff
@@ -65,6 +65,8 @@ var (
 		Version:  hcVersion,
 		Resource: hcResourcePlural,
 	}
+
+	defaultAcct = "notdemo"
```
tnthornton, line 69: This is to make it obvious that something didn't go as expected. In general we shouldn't see this account specifically.

cmd/up/controlplane/connect.go
```diff
@@ -0,0 +1,202 @@
...
+	if err != nil {
+		return err
+	}
+
+	modifiedCfg, err := updateKubeConfig(*cfg, upCtx.Account, c.Name, origContext)
+	if err != nil {
+		return err
+	}
+	// NOTE(tnthornton) we don't current support supplying files outside of
+	// the default system kubeconfig.
```
darkmuggle, line 146: 😢 that's a bit unfortunate

darkmuggle (approved):
A few trivial nits, but overall LGTM

cmd/up/controlplane/connect.go
tnthornton: I did that due to `up ctp connect one --kubeconfig=/path` and `up ctp disconnect --kubeconfig=/path`, seemed like an uncommon case.

We can add it in the future if it turns out I'm mistaken.

cmd/up/controlplane/disconnect.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func switchContext(cfg api.Config, target string) error {
+	cfg.CurrentContext = target
+	return clientcmd.ModifyConfig(clientcmd.NewDefaultClientConfigLoadingRules(), cfg, false)
+}
+
+func removeFromConfig(cfg api.Config, contextName string) (api.Config, error) {
+	if cfg.CurrentContext == contextName {
+		return api.Config{}, fmt.Errorf(errFmtCurrentContext, contextName)
```
AlainRoy, line 98: In your demo, you did a connect immediately followed by a disconnect. I would have expected this error message (based on the code) but didn't see it. Am I just misreading the code?

For what it's worth, I thought the demo flow was right -- I didn't see a need for this error.

AlainRoy (approved):
Looks good to me! One question in the review. 

One other question: did you also test with cloud control planes to ensure they work? They demo only showed spaces.

tnthornton:
>One other question: did you also test with cloud control planes to ensure they work? They demo only showed spaces.

>3. Verified functionality was consistent between spaces and cloud.

cmd/up/controlplane/disconnect.go
tnthornton: This https://github.com/upbound/up/pull/393/files#diff-ec80f85a9240e92b2ee2058fb4a060dc60969d7f45e0e3801a3e50dc56ac67f0R70 switches the context to the original context prior to the removal.

The error itself is to help someone utilizing this function to know they are doing something wrong.

cmd/up/controlplane/connect.go
tnthornton: Also to clarify, if you set the `KUBECONFIG` environment variable that counts as "the default system kubeconfig".

cmd/up/controlplane/disconnect.go
AlainRoy: Cool, thanks.

cmd/up/controlplane/connect.go
sttts: Good news is that we can reuse the token when accessing another control plane (after https://github.com/upbound/up/issues/401 is fixed). So this is just once per user.

What we certainly could do is automating this by calling the API token API, at least after some interactive `[Ny]` question to the user.

cmd/up/controlplane/controlplane.go
```diff
@@ -88,10 +88,12 @@ func PredictControlPlanes() complete.Predictor {
 
 // Cmd contains commands for interacting with control planes.
 type Cmd struct {
-	Create createCmd `cmd:"" help:"Create a managed control plane."`
-	Delete deleteCmd `cmd:"" help:"Delete a control plane."`
-	List   listCmd   `cmd:"" help:"List control planes for the account."`
-	Get    getCmd    `cmd:"" help:"Get a single control plane."`
+	Connect    connectCmd    `cmd:"" help:"Connect kubectl to control plane."`
+	Disconnect disconnectCmd `cmd:"" help:"Disconnect kubectl from control plane."`
```
sttts, line 92: This is a pretty strange UX with its profile handling magic, changing back to the previous context via disconnect, and error'ing if meanwhile one wanted to talk to some other kube cluster.

PR #421, opened 2024-01-31
add alpha command for web login

nullable-eth:
Create alpha command for `up alpha web-login`.  This supports SSO integration.  When enabling SSO, upbound-api locks logins to SSO only and this means the cli would no longer be able to be used by those orgs.  This allows web and SSO based logins via the `up` cli by creating a server listening for a callback, it then opens a browser to do a login.  Any type of login can be done here, password, social or SSO (if SSO is enabled on an org then ONLY SSO may be used) and once logged in redirects to the callback in `up` cli passing it a time based one-time password.  This TOTP is then exchanged for a session in upbound-api  which is then stored and used by the `up` cli.

Note: in the case of headless terminals (ssh in to server) where a graphical web-browser cannot be used to complete SSO the `up` cli detects this and displays a link and QR code that can be used to authenticate on your phone or other device, once authenticated you can type in the 6 digit TOTP code and `up` cli will complete the login.

Fixes #

I have:

Tested locally and against dev


cmd/up/login/web.go
```diff
@@ -0,0 +1,314 @@
+// Copyright 2022 Upbound Inc
```
AlainRoy, line 1: Nit: 2024

cmd/up/login/web.go
```diff
@@ -0,0 +1,314 @@
...
+	return nil
+}
+
+func (c *LoginWebCmd) AfterApply(kongCtx *kong.Context) error {
+	upCtx, err := upbound.NewFromFlags(c.Flags, upbound.AllowMissingProfile())
+	if err != nil {
+		return err
+	}
+	// NOTE(hasheddan): client timeout is handled with request context.
+	// TODO(hasheddan): we can't use the typical up-sdk-go client here because
```
AlainRoy, line 66: hasheddan in new code?

cmd/up/login/web.go
```diff
@@ -0,0 +1,314 @@
...
+			"message": []string{err.Error()},
+		}.Encode()
+		redirect <- resultEP.String()
+		return errors.Wrap(err, errLoginFailed)
+	}
+	redirect <- resultEP.String()
+	return nil
+}
+
+func openBrowser(url string) error {
```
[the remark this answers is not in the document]
nullable-eth: Yes, I used that package originally.  However on errors from the browser failing to open in headless sessions, there was no way to override the errors being printed to stdout which just muddied the display.  I suppose I could just globally set the writers to nil on the whole package level but initially that seemed heavy handed but I'll test it out.

cmd/up/login/web.go
nullable-eth: Code copied from the login method and the comments still valid, plus will always preserve @hasheddan.  All hail Dan!

cmd/up/login/web.go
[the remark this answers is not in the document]
nullable-eth: Overriding the package level writers worked, so I went back to using the browser package.

branden (approved):

PR #422, opened 2024-01-31
Add step to wait for PackageRevisions to be healthy

tnthornton:
We've had some internal reports of migration not working, with what appears to be a race between attempting to apply resources prior to the APIs being available. This changeset adds a step to wait for the various PackageRevisions to be healthy before moving on.

I have:

### DIFF
--- a/internal/migration/importer/import.go
+++ b/internal/migration/importer/import.go
@@ -176,6 +176,20 @@ func (im *ControlPlaneStateImporter) Import(ctx context.Context) error { // noli
 		}
 	}
 	s.Success(waitPkgsMsg + "Installed and Healthy! ⏳")
+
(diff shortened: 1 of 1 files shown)


haarchri (approved):
thanks for this enhancements ;) it's very missleading that Providers,Functions, Configurations Healthy / Installed when the Revisions Not in this Status ..

turkenh:
Thanks for the PR @tnthornton 🙌 

> it's very missleading that Providers,Functions, Configurations Healthy / Installed when the Revisions Not in this Status ..

I couldn't reproduce this with Crossplane v1.14 but able to reproduce with Crossplane 1.13. 
So, the good news is this is something already fixed in the latest version presumably with [this PR](https://github.com/crossplane/crossplane/pull/4647).

PR #426, opened 2024-02-05
Do not restore package statuses and use the same confirmation dialog as spaces init

turkenh:

This PR is a follow up on the problem described in https://github.com/upbound/up/pull/422 during import:

- Do not restore status for base resources
  - Otherwise, packages reported as installed/healthy before they were actually processed. 
  - No need for restoring status for any other base resources, so, disabled all together.
- Unifies the confirmation dialog as `spaces init` 
  - To confirm exporting secrets.
  - When preflight checks failed during import.

I have:

1. Create a control plane
2. Stop Crossplane
3. Import
4. Ensure it waits successfully.

### DIFF
--- a/cmd/up/migration/export.go
+++ b/cmd/up/migration/export.go
@@ -16,8 +16,8 @@ package migration
 
 import (
 	"context"
-	"fmt"
 
+	"github.com/pterm/pterm"
 	apiextensionsclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/discovery/cached/memory"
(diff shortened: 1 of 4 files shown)


cmd/up/migration/export.go
```diff
@@ -37,7 +37,7 @@ reconstructable post-migration. Consequently, the exported archive will incorpor
 those secrets by default. To exclude secrets from the export, please use the
 --excluded-resources flag.
 
-IMPORTANT: The exported archive will contain secrets. Do you wish to proceed? [y/n]`
+IMPORTANT: The exported archive will contain secrets. Do you wish to proceed?`
```
tnthornton, line 40: Is anything else presented to the user for how to proceed if we remove [y/n]?

turkenh: Yes, this is how it looks like: 

Screenshot 2024-02-06 at 11 20 44

tnthornton (approved):

Issue #401, opened 2023-10-26
`up ctp connect` works exactly once


sttts:
### What happened?

`up ctp connect` (added in https://github.com/upbound/up/pull/393/files) works with the first control plane. When trying to run it again to switch to another control plane, this code triggers and the command does nothing:
```golang
	// Check if the fs kubeconfig is already pointing to a control plane and
	// return early if so.
	origContext := kcloader.CurrentContext
	if strings.HasPrefix(origContext, upboundPrefix) {
		return nil
	}
```

### How can we reproduce it?

Run `up ctp connect` against two control planes. Second command is a noop.

### What environment did it happen in?

Cloud prod.

sttts:
I guess the intention was that `up ctp disconnect` restored the old context, and before doing that `up ctp connect` is a noop. That surprising for the user and not documented. 

Am not a fan of that, but at least it should print `not doing anything because a controlplane is already connected` and return with an error. It would be more logical to just connect to the new controlplane instead. It would be what the command says: connect.

AlainRoy:
I've assigned it to @tnthornton to fix next week. He's the one that implemented it and he can comment on the intended behavior. I agree -- I also find the behavior surprising, but we can discuss it with him next week instead of preemptively changing it. He'll be back on Tuesday.

sttts:
A related topic here is about tokens. The `--token` flag kind of destroys the experience. If we reused the token from a previous connect, that would make this much nicer. E.g. by defining an auth setting in the kubeconfig as the default one for Upbound cloud.

sttts:
Also in this context, maybe I just missed it and it already exists: `--context <the-kube-context-name-this-ctp`. People use kubectx or other tooling to switch between Kube clusters all the time. This would make Upbound compatible with those workflows.

AlainRoy:
Taylor and I chatted about tokens before -- is there a way we can eliminate the need for the tokens? Most other commands don't require a token. Why should this one/

tnthornton:
>I guess the intention was that up ctp disconnect restored the old context, and before doing that up ctp connect is a noop. That's surprising for the user and not documented.

That is accurate. Originally the idea was that you were only ever attempting to connect from a "real" cluster; hence the need to disconnect before attempt to connect to a different control plane.

[...]

PR #429, opened 2024-02-11
Improve UX of ctp connect+disconnect

sttts:
What is chagned:

1. print what connect and disconnect do, aka switch contexts in the kubeconfig.
2. fix handling of contexts with underscore. This had been broken on disconnect.

Fixes https://github.com/upbound/up/issues/401.

### DIFF
--- a/cmd/up/controlplane/connect.go
+++ b/cmd/up/controlplane/connect.go
@@ -131,7 +131,7 @@ func (c *connectCmd) Run(ctx context.Context, printer upterm.ObjectPrinter, p pt
 	// return early if so.
 	origContext := kcloader.CurrentContext
 	if strings.HasPrefix(origContext, upboundPrefix) {
-		return nil
+		return fmt.Errorf("already connected to a control plane. Disconnect first")
 	}
 
 	cfg, err := c.client.GetKubeConfig(ctx, types.NamespacedName{Namespace: c.Group, Name: c.Name})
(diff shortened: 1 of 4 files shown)


cmd/up/controlplane/connect.go
```diff
@@ -131,7 +131,7 @@ func (c *connectCmd) Run(ctx context.Context, printer upterm.ObjectPrinter, p pt
 	// return early if so.
 	origContext := kcloader.CurrentContext
 	if strings.HasPrefix(origContext, upboundPrefix) {
-		return nil
+		return fmt.Errorf("already connected to a control plane. Disconnect first")
```
AlainRoy, line 134: I appreciate you're keeping the same behavior, but with a better error message. But I wonder if doing the second "connect" should automatically disconnect from the first control plane? An idea for the future, not a blocker for this PR.

AlainRoy (approved):
Looks good to me!

cmd/up/controlplane/connect.go
sttts: I agree. That would be even better. Let me see if it is easy to do.

sttts: It's not that obvious. Needs a follow-up.

tnthornton: >It's not that obvious. Needs a follow-up.

FWIW - that's why it didn't land with that behavior the first time. It's not entirely obvious how to do that at this point.

sttts: Not sure I see fundamental problems. But I might miss something. Got lost more in the kubeconfig massaging code that is full of side-effects.

PR #431, opened 2024-02-27
connect: gracefully handle different context names

### LINKED ISSUE #407: `up ctp connect` does not work with spaces
### What happened?

Two issues:
1. even though on a Space profile, the `up ctp connect` command requires an account (which is not set for a pure-Spaces profile):

   ```
   up: error: controlplane.connectCmd.Run(): error: account is missing from profile
   ```
3. the connection secret holding the kubeconfig of a Space controlplane has arbitrary cluster, auth and context names. The `up ctp connect` code requires a format used by the cloud:

   ```
   up: error: controlplane.connectCmd.Run(): config is broken, missing cluster: "somefakeaccount-ctp1"
   ```

### How can we reproduce it?

Create a profile with `up profile set space --profile space` and switch to it.

Then `up ctp connect` to some controlplane on that Spaces cluster.

### What environment did it happen in?

All latest main branches.

sttts:
Fixes #407.

### DIFF
--- a/cmd/up/controlplane/connect.go
+++ b/cmd/up/controlplane/connect.go
@@ -143,7 +143,9 @@ func (c *connectCmd) Run(ctx context.Context, printer upterm.ObjectPrinter, p pt
 		return err
 	}
 
-	modifiedCfg, err := updateKubeConfig(*cfg, upCtx.Account, c.Name, origContext)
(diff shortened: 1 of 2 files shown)


AlainRoy (approved):
Looks good to me!

PR #433, opened 2024-02-27
Unify 'up ctp connect' and 'up ctp kubeconfig get'

sttts:
Use the same code for both commands, and make `up ctp kubeconfig get` work with spaces.

Signed-off-by: Dr. Stefan Schimanski <stefan.schimanski@upbound.com>

### DIFF
--- a/cmd/up/controlplane/connect.go
+++ b/cmd/up/controlplane/connect.go
@@ -18,103 +18,36 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io"
-	"os"
 	"strings"
 
 	"github.com/alecthomas/kong"
 	"github.com/pterm/pterm"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/tools/clientcmd"
-	"k8s.io/client-go/tools/clientcmd/api"
 
-	"github.com/upbound/up-sdk-go/service/configurations"
-	cp "github.com/upbound/up-sdk-go/service/controlplanes"
+	"github.com/upbound/up/cmd/up/controlplane/kubeconfig"
 	"github.com/upbound/up/internal/controlplane"
-	"github.com/upbound/up/internal/controlplane/cloud"
-	"github.com/upbound/up/internal/controlplane/space"
 	"github.com/upbound/up/internal/kube"
 	"github.com/upbound/up/internal/upbound"
-	"github.com/upbound/up/internal/upterm"
 )
 
 const (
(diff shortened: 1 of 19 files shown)


cmd/up/controlplane/kubeconfig/connection.go
```diff
@@ -0,0 +1,147 @@
...
+	errFmtConfigBroken = "config is broken, missing %s: %q"
+)
+
+// ConnectionSecretCmd is the base for command getting connection secret for a control plane.
+type ConnectionSecretCmd struct {
+	Name  string `arg:"" required:"" help:"Name of control plane." predictor:"ctps"`
+	Token string `help:"API token used to authenticate. Required for Upbound Cloud; ignored otherwise."`
+	Group string `short:"g" help:"The control plane group that the control plane is contained in. By default, this is the group specified in the current profile."`
+
+	stdin io.Reader
```
AlainRoy, line 45: Do we need this? It's always set to os.stdin -- is it a useful level of indirection?

cmd/up/controlplane/connect.go
```diff
@@ -127,83 +60,43 @@ func (c *connectCmd) Run(ctx context.Context, printer upterm.ObjectPrinter, p pt
...
-	}
-	cfg.AuthInfos[ctpKey] = users
-	delete(cfg.AuthInfos, sourceKey)
-
-	// Rename context, move under upbound- namespace.
-	context, ok := cfg.Contexts[sourceKey]
-	if !ok {
-		return api.Config{}, fmt.Errorf(errFmtConfigBroken, "context", sourceKey)
+func controlplaneContextName(account string, name types.NamespacedName, origCtx string) string {
+	if name.Namespace == "" {
```
AlainRoy, line 98: Personal preference (but I know you're a way better and more idiomatic Go developer than me, so you probably should ignore me): Because the sprintf() is a bit complicated, I'd rather see it occur just once. Something like:

namespace := name.Namespace
if namespace == "" {
  namespace = "default"
}
return fm.Sprintf(...

AlainRoy (approved):
Looks good to me! Two small nits, but they're non-blocking.

cmd/up/controlplane/kubeconfig/connection.go
sttts: good point. Blindly cargo culted.

sttts: fixed

cmd/up/controlplane/connect.go
sttts: fixed

PR #434, opened 2024-02-28
Allow repeated ctp connects

sttts:
Do the full fix of #401.


cmd/up/controlplane/connect.go
```diff
@@ -62,9 +62,18 @@ func (c *connectCmd) Run(ctx context.Context, p pterm.TextPrinter, upCtx *upboun
 	}
 
 	// Check if the fs kubeconfig is already pointing to a control plane and return early if so.
```
AlainRoy, line 64: Nit: I think this comment is now wrong.

cmd/up/controlplane/connect.go
```diff
@@ -62,9 +62,18 @@ func (c *connectCmd) Run(ctx context.Context, p pterm.TextPrinter, upCtx *upboun
...
+		// disconnect first
+		kubeConfig, err = switchToOrigContext(kubeConfig)
+		if err != nil {
+			return fmt.Errorf("context %q seems to be a control plane context, but disconnect failed: %w", oldContext, err)
+		}
+		oldContext = kubeConfig.CurrentContext
+		if err := clientcmd.ModifyConfig(clientcmd.NewDefaultClientConfigLoadingRules(), kubeConfig, false); err != nil {
+			return err
+		}
+		p.Printfln("Switched back to context %q.", oldContext)
```
AlainRoy, line 76: I wonder if this will be confusing for a user. If I'm reading this correctly, they'll see two messages in a row:

> Switched back to context <old-context>
> Current context set to <new-context>

That looks confusing to me. Do you agree?

AlainRoy (approved):
One nit, one question about how this looks to a user (not sure about my question). Code looks okay, so I'll approve. If my question makes sense, I'll trust you to fix that bit without a second review. :)

cmd/up/controlplane/connect.go
sttts: done

cmd/up/controlplane/connect.go
sttts: fixed. Not 100% sure which is better though.

PR #439, opened 2024-03-11
feat: move export to dedicated module

phisco:

This PR:
- Moves the migration code to a dedicated Go module
- Reimplements its archive related code to avoid importing `github.com/mholt/archiver/v4`
- Makes its spinner pluggable, defaulting to a no-op

I have:

- [ ] ~Added `backport release-x.y` labels to auto-backport this PR, as appropriate.~

The output of an export using the code from main:
image
The output of an export using the code from this PR:
image

The diff between the two generated archives using `pkgdiff foo.tar.gz main.tar.gz`, returned only timestamps differences.

Import:
```
$ kubectl create secret generic super-secret-secret --from-literal=password=supersecret
$ go run github.com/upbound/up/cmd/up alpha migration export --output foo.tar.gz
$ go run github.com/upbound/up/cmd/up alpha migration import -i foo.tar.gz
$ k get secret super-secret-password -o yaml
```


pkg/migration/exporter/export.go
```diff
@@ -331,16 +323,84 @@ func (e *ControlPlaneStateExporter) archive(ctx context.Context, fs afero.Afero,
...
+		if _, err := io.Copy(tw, f); err != nil {
+			return err
+		}
+
+		return nil
+	})
 
-	return format.Archive(ctx, out, files)
+	// Return any errors encountered while creating the archive
+	return err
```
turkenh, line 403: suggestion: it would be helpful to wrap the errors with more context.

pkg/migration/go.mod
```diff
@@ -0,0 +1,80 @@
...
+	google.golang.org/appengine v1.6.7 // indirect
+	google.golang.org/protobuf v1.32.0 // indirect
+	gopkg.in/inf.v0 v0.9.1 // indirect
+	k8s.io/klog/v2 v2.100.1 // indirect
+	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
+)
+
+replace (
+	github.com/crossplane/crossplane/controller/apiextensions => ../../internal/vendor/github.com/crossplane/crossplane/controller/apiextensions
```
turkenh, line 76: why we need thinks, can we avoid 🤔

turkenh (approved):
Great work @phisco, looking good 🙌

pkg/migration/exporter/export.go
phisco: updated 🙏

pkg/migration/go.mod
phisco: not sure what we changed in the vendored version tbh 😓

PR #445, opened 2024-03-14
WIP: Add `up alpha query` command

sttts:


cmd/up/query/native.go
```diff
@@ -0,0 +1,101 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package query
+
+var (
+	nativeKinds     = map[string]string{}
+	nativeResources = map[string]string{
```
tnthornton, line 19: Do we need these? I'm only seeing them referenced in this file and the corresponding test so far.

cmd/up/query/resource/query_types.go
```diff
@@ -0,0 +1,137 @@
+// Copyright 2024 Upbound Inc.
+// All rights reserved
+
+package resource
```
tnthornton, line 4: Why does this file need to live here instead of in up-sdk-go? Seems like it's coupled to the query API - but maybe I'm missing something 😅

cmd/up/main.go
```diff
@@ -145,14 +146,6 @@ func main() {
...
 		kong.Name("up"),
 		kong.Description("The Upbound CLI"),
-		kong.Help(func(options kong.HelpOptions, ctx *kong.Context) error {
-			// Do not emit help if command is hidden.
-			if ctx.Selected() != nil && ctx.Selected().Hidden {
-				fmt.Fprintf(ctx.Stdout, "Refusing to emit help for hidden command. See %s variant.\n", feature.GetMaturity(ctx.Selected()))
-				return nil
-			}
-			return kong.DefaultHelpPrinter(options, ctx)
-		}),
```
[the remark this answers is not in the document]
sttts: Why should we hide help in hidden commands in an opensource project?

cmd/up/query/args.go
```diff
@@ -0,0 +1,204 @@
...
+var nativeScheme = runtime.NewScheme()
+
+func init() {
+	kruntime.Must(apiextensionsv1.AddToScheme(nativeScheme))
+	kruntime.Must(apiextensionsv1beta1.AddToScheme(nativeScheme))
+	kruntime.Must(clientgoscheme.AddToScheme(nativeScheme))
+}
+
+// This file mostly comes from k8s.io/cli-runtime/pkg/resource, adapted for our
+// use-cases without the builder.
```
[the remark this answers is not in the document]
sttts: It's copied and modified. Maybe 80% is still the old code. It's not exposed by Kube, but also not super complicated. So I think forking it here is fine.

cmd/up/query/native.go
sttts: Typical helper pattern. I believe Crossplane uses the same. It's not a good pattern to share utilitites, hence I keep them rather here.

cmd/up/query/resource/query_types.go
sttts: Typical helper pattern. I believe Crossplane uses the same. It's not a good pattern to share utilitites, hence I keep them rather here.

cmd/up/main.go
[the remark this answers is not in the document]
tnthornton: I think the original intent was if it's hidden, there shouldn't be any output (which is arguably a UX limitation).

cmd/up/query/native.go
tnthornton: >It's used to map resources to kinds in args.go.

Gotcha. Must've just missed it 👍

cmd/up/main.go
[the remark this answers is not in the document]
sttts: There is only output if you do `up hidden --help`

cmd/up/query/native.go
sttts: For reference. Same code in https://github.com/upbound/spaces/pull/717.

cmd/up/query/resource/query_types.go
tnthornton: Understood. I was mostly skimming this in the context of https://github.com/upbound/up-sdk-go/blob/5a36ec6e89864282c636203a9f5f33d8a6b5c7f8/apis/query/v1alpha1/queryresponse_types.go which seemed duplicative. Looking at what we have here though, this is different.

cmd/up/main.go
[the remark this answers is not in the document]
tnthornton: 👍

tnthornton (approved):

cmd/up/query/resource/query_types.go
sttts: Oh, the behaviour is slight different. If you inherit `DeepCopyObject` from an embedded type you actually get back the embedded type, not the outer one you actually want. And the other deepcopy funcs are slightly different in type.

Actually, I was not interested in the deepcopy funcs, but in the setters and getters to enable generic code for all 3 scopes.

PR #446, opened 2024-03-15
Align trace's arguments with query+get and fix up/down navigation

sttts:
Use the same positional argument code and resource type parsing from query and make trace use it. So you can write the same expression as in `kubectl get`, e.g.
```shell
	# Trace all buckets.
	up alpha trace buckets
        
	# Trace all Crossplane claims.
	up alpha trace claims

	# Trace buckets and vpcs.
	up alpha trace buckets,vpc

	# Trace the buckets prod and staging.
	up alpha trace buckets prod staging

	# Trace the bucket prod and the vpc default.
	up alpha trace bucket/prod vpc/default 
```


cmd/up/query/resource/query_types.go
```diff
@@ -78,13 +78,15 @@ func (q *SpaceQuery) GetSpec() *queryv1alpha1.QuerySpec {
...
 // SetSpec sets the spec of the query.
-func (q *SpaceQuery) SetSpec(spec *queryv1alpha1.QuerySpec) {
+func (q *SpaceQuery) SetSpec(spec *queryv1alpha1.QuerySpec) QueryObject {
 	q.Spec = spec
+	return q
 }
 
 // SetResponse sets the response of the query.
-func (q *SpaceQuery) SetResponse(response *queryv1alpha1.QueryResponse) {
+func (q *SpaceQuery) SetResponse(response *queryv1alpha1.QueryResponse) QueryObject {
```
tnthornton, line 87: OK - silly question time. You have a pointer receiver that you're mutating, why do we need to return the object?

Edit --
Maybe I answered my question. I see this below:
```golang
			query := queryObject.DeepCopyQueryObject().SetSpec(spec)
			if err := kc.Create(ctx, query); err != nil {
				return nil, fmt.Errorf("%T request failed: %w", query, err)
			}
			objs = append(objs, query.GetResponse().Objects...)
```
So rather than getting the query object and then doing a SetSpec(), we're folding it into 1 line 🤔. Well 1 line instead of 3 - get object, set spec, use object.

cmd/up/trace/app.go
```diff
@@ -114,11 +116,12 @@ func NewApp(title string, group, kind, name string, pollFn func(kind, group, nam
 		).
 		SetSubTitles(upviews.GridTitle{Col: 0, Row: 2, Fn: func(screen tcell.Screen, x, y, w int) {
 			err := app.model.Error.Load().(string)
-			tview.Print(screen, err, x, y, w, tview.AlignCenter, tcell.ColorHotPink)
+			tview.Print(screen, err, x, y-1, w, tview.AlignCenter, tcell.ColorHotPink)
```
tnthornton, line 119: What's magic about the y-1 vs y? Small nit, could we add a short comment about why we need the -1?

tnthornton (approved):
Just have the small nit for a comment on why `y-1` vs `y` 👍

cmd/up/query/resource/query_types.go
sttts: Builder pattern basically. `SetSpec` versus `WithSpec`. `Set*` has side-effects.

cmd/up/trace/app.go
sttts: Oh, in `y` is the F-key bar that is drawn above. We could add an error bar on-top of that. Maybe I do that, with flashy red background.

sttts: Fixed

PR #457, opened 2024-04-02
Make ctp list derive context from kubeconfig

Piotr1215:

This PR makes `up ctp list` kubeconfig context aware, making it independent from the current profile. The profile is automatically derived:

$ kubectl ctx not-space-kube
$ up ctp list
You are not on a space.
$ kubectl ctx space
$ up ctp list
.... controlplanes from the profile matching the kubeconfig context's URL ....
$ up ctp connect ctp
$ up ctp list
In a controlplane, you cannot list controlplanes.
Use 'up ctx ..' to go to group level.

Fixes https://github.com/upbound/spaces/issues/752

NOTE: this breaks the use against a cloud profile. We will bring it back in a follow-up.

I have:

- [ ] ~Added `backport release-x.y` labels to auto-backport this PR, as appropriate.~

This can be tested locally by running the following setup:

- create local kind cluster with spaces
- create empty kind cluster
- follow the command logic from description

### DIFF
--- a/cmd/up/controlplane/list.go
+++ b/cmd/up/controlplane/list.go
@@ -19,82 +19,68 @@ import (
 
 	"github.com/alecthomas/kong"
 	"github.com/pterm/pterm"
-	"k8s.io/client-go/dynamic"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 
-	"github.com/upbound/up-sdk-go/service/configurations"
-	cp "github.com/upbound/up-sdk-go/service/controlplanes"
-	"github.com/upbound/up/internal/controlplane"
-	"github.com/upbound/up/internal/controlplane/cloud"
-	"github.com/upbound/up/internal/controlplane/space"
+	"github.com/crossplane/crossplane-runtime/pkg/errors"
+
+	spacesv1beta1 "github.com/upbound/up-sdk-go/apis/spaces/v1beta1"
+	"github.com/upbound/up/internal/profile"
 	"github.com/upbound/up/internal/upbound"
 	"github.com/upbound/up/internal/upterm"
 )
-
-type ctpLister interface {
-	List(ctx context.Context, namespace string) ([]*controlplane.Response, error)
-}
 
 // listCmd list control planes in an account on Upbound.
 type listCmd struct {
 	Group     string `short:"g" help:"The control plane group that the control plane is contained in. This defaults to the group specified in the current profile."`
 	AllGroups bool   `short:"A" default:"false" help:"List control planes across all groups."`
-
-	client ctpLister
 }
 
 // AfterApply sets default values in command after assignment and validation.
-func (c *listCmd) AfterApply(kongCtx *kong.Context, upCtx *upbound.Context) error {
-	if upCtx.Profile.IsSpace() {
-		kubeconfig, ns, err := upCtx.Profile.GetSpaceRestConfig()
-		if err != nil {
-			return err
-		}
-		if c.Group == "" {
-			c.Group = ns
-		}
-
-		client, err := dynamic.NewForConfig(kubeconfig)
-		if err != nil {
-			return err
-		}
-		c.client = space.New(client)
-	} else {
-		cfg, err := upCtx.BuildSDKConfig()
-		if err != nil {
-			return err
-		}
-		ctpclient := cp.NewClient(cfg)
-		cfgclient := configurations.NewClient(cfg)
-
-		c.client = cloud.New(ctpclient, cfgclient, upCtx.Account)
-	}
-
+func (c *listCmd) AfterApply(kongCtx *kong.Context) error {
 	kongCtx.Bind(pterm.DefaultTable.WithWriter(kongCtx.Stdout).WithSeparator("   "))
+
 	return nil
 }
 
 // Run executes the list command.
-func (c *listCmd) Run(ctx context.Context, printer upterm.ObjectPrinter, p pterm.TextPrinter, upCtx *upbound.Context) error {
-	l, err := c.client.List(ctx, c.deriveGroup())
-	if controlplane.IsNotFound(err) {
-		p.Printfln("No Control planes found in %s group", c.deriveGroup())
-		return nil
+func (c *listCmd) Run(ctx context.Context, printer upterm.ObjectPrinter, upCtx *upbound.Context, p pterm.TextPrinter) error { // nolint:gocyclo
+	// get context
+	_, currentProfile, ctp, err := upCtx.Cfg.GetCurrentContext(ctx)
+	if err != nil {
+		return err
 	}
+	if currentProfile == nil {
+		return errors.New(profile.NoSpacesContextMsg)
+	}
+	if ctp.Namespace == "" {
+		return errors.New(profile.NoGroupMsg)
+	}
+	if ctp.Name != "" {
+		return errors.New("Cannot list control planes from inside a control plane, use `up ctx ..` to switch to a group level.")
+	}
+
+	// create client
+	restConfig, _, err := currentProfile.GetSpaceRestConfig()
+	if err != nil {
+		return err
+	}
+	cl, err := ctrlclient.New(restConfig, ctrlclient.Options{})
 	if err != nil {
 		return err
 	}
 
-	if len(l) == 0 {
-		p.Println("No control planes found")
-		return nil
+	// list control planes
+	ns := ctp.Namespace
+	if c.Group != "" {
+		ns = c.Group
 	}
-
-	return tabularPrint(l, printer, upCtx)
-}
-
-func (c *listCmd) deriveGroup() string {
 	if c.AllGroups {
-		return ""
+		ns = ""
 	}
-	return c.Group
+	var ctps spacesv1beta1.ControlPlaneList
+	if err := cl.List(ctx, &ctps, ctrlclient.InNamespace(ns)); err != nil {
+		return err
+	}
+
+	return printer.Print(ctps.Items, spacefieldNames, extractSpaceFields)
 }
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -16,10 +16,14 @@ package config
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"io"
 	"os"
 	"path/filepath"
+
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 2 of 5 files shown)


cmd/up/controlplane/list.go
```diff
@@ -75,7 +77,28 @@ func (c *listCmd) AfterApply(kongCtx *kong.Context, upCtx *upbound.Context) erro
...
+
+	if err != nil {
+		return err
+	}
+	if p == nil {
+		return nil
+	}
+
 	l, err := c.client.List(ctx, c.deriveGroup())
+
```
sttts, line 101: same

cmd/up/controlplane/list.go
```diff
@@ -89,6 +112,12 @@ func (c *listCmd) Run(ctx context.Context, printer upterm.ObjectPrinter, p pterm
 		return nil
 	}
 
+	for _, resp := range l {
+		if ctp.Name == resp.Name {
+		p.Printfln("Cannot list control from inside a controlplane, use `up ctx ..` to switch to a goup level")
+		return nil
+		}
+	}
```
sttts, line 120: not sure I understand this for loop? Why not check for `ctp.Name == ""` at line 91?

cmd/up/controlplane/list.go
```diff
@@ -45,7 +47,7 @@ type listCmd struct {
 // AfterApply sets default values in command after assignment and validation.
 func (c *listCmd) AfterApply(kongCtx *kong.Context, upCtx *upbound.Context) error {
 	if upCtx.Profile.IsSpace() {
-		kubeconfig, ns, err := upCtx.Profile.GetSpaceKubeConfig()
+		kubeconfig, ns, err := upCtx.Profile.GetSpaceRestConfig()
```
sttts, line 50: actually I would expect the logic about the current context here. I.e. either move the client creation down into Run or move the `profile.FromKubeconfig` call here. We should not create a client with invalid rest config, and especially not try to list.

sttts: Alternative: change `GetSpaceKubeConfig` to call `profile.FromKubeconfig`. That will cover all commands that rely on being in the space.

cmd/up/controlplane/list.go
Piotr1215: I thought I had linter configured on `git commit hook` but not for this project. Should be fixed now.

cmd/up/controlplane/list.go
Piotr1215: Because this is an array `var l []*controlplane.Response` so we could be in any the control planes listed.

cmd/up/controlplane/list.go
```diff
@@ -75,7 +77,28 @@ func (c *listCmd) AfterApply(kongCtx *kong.Context, upCtx *upbound.Context) erro
...
+	if err != nil {
+		return err
+	}
+
+	_, _, ctp, err := profile.FromKubeconfig(ctx, profiles, conf)
+
+	if err != nil {
+		return err
+	}
+	if p == nil {
```
tnthornton, line 97: I believe `p` is supplied by Kong. Is there a case where this would be nil?

cmd/up/controlplane/list.go
Piotr1215: @sttts moved everything out of the `AfterApply` method to `Run` so that the client creation is in this method now.

justfile
```diff
@@ -0,0 +1,30 @@
...
+  cd /home/decoder/dev/mxe
+  make cluster.up tilt.up
+
+# setup empty kind cluster for testing
+setup_empty_kind:
+  kind create cluster --name empty
+
+# create test control plane
+create_control_plane:
+  just "up ctp create test-ctp"
```
sttts, line 30: how is this set of commands easier than just calling the commands directly?

cmd/up/controlplane/list.go
sttts: The client is still a group client and we list even though we are in a controlplane. Could well be that we have to further separate the cloud code from the spaces code. But how it is now might give the desired output, but it's not the straight forward way to do it. I.e. we list and retroactively check whether the list contains something that resembles our current context. Actually it might even be wrong because the list might come from a wrong profile.

justfile
Piotr1215: Oups, sorry this shouldn't have been checked in. It's _just_ for my dev workflow.

Piotr1215: Removed from the repo

cmd/up/controlplane/list.go
Piotr1215: @sttts I _think_ I can follow the logic now. Some of the code is duplicated in the _if_ branches, I will fix it later, just wanted to check if I'm on the right track with the implementation.

cmd/up/controlplane/list.go
```diff
@@ -58,6 +86,28 @@ func (c *listCmd) AfterApply(kongCtx *kong.Context, upCtx *upbound.Context) erro
 			return err
 		}
 		c.client = space.New(client)
+		l, err := c.client.List(ctx, c.deriveGroup())
+
+		if controlplane.IsNotFound(err) {
```
sttts, line 91: this cannot happen

cmd/up/controlplane/list.go
```diff
@@ -44,20 +47,49 @@ type listCmd struct {
...
+		}
+		profiles, err := upCtx.Cfg.GetUpboundProfiles()
+		if err != nil {
+			return err
+		}
+		if err != nil {
+			return err
+		}
+
+		_, curprof, ctp, err := profile.FromKubeconfig(ctx, profiles, conf)
```
sttts, line 70: Dieser ganze Block wird häufig auftreten. Wir sollten einen Helper dazu erstellen:
```
profileName, profile, ctp, restConfig, err := upCtx.Cfg.GetCurrentContext(ctx)
```

cmd/up/controlplane/list.go
```diff
@@ -44,20 +44,40 @@ type listCmd struct {
...
-		if c.Group == "" {
-			c.Group = ns
+
+		// if ctp.Name is populated this means we are at the control plane level
+		if ctp.Name != "" {
+			p.Printfln("Cannot list control planes from inside a control plane, use `up ctx ..` to switch to a group level")
+			return nil
+		}
+
+		if err != nil {
```
sttts, line 66:
```suggestion
		if err != nil {
```

sttts: And move the error handling above the name check.

cmd/up/controlplane/list.go
```diff
@@ -44,20 +44,40 @@ type listCmd struct {
...
+	kongCtx.Bind(pterm.DefaultTable.WithWriter(kongCtx.Stdout).WithSeparator("   "))
+	return nil
+}
+
+// Run executes the list command.
+func (c *listCmd) Run(ctx context.Context, printer upterm.ObjectPrinter, p pterm.TextPrinter, upCtx *upbound.Context) error { // nolint:gocyclo
 	if upCtx.Profile.IsSpace() {
-		kubeconfig, ns, err := upCtx.Profile.GetSpaceKubeConfig()
+		_, curprof, ctp, err := upCtx.Cfg.GetCurrentContext(ctx)
+
```
sttts, line 55: Never an empty line before the error handling.

internal/config/config.go
```diff
@@ -122,6 +125,27 @@ func (c *Config) GetUpboundProfile(name string) (profile.Profile, error) {
 	return p, nil
 }
 
+// GetCurrentContext returns the current context from the kubeconfig, profile and config
+func (c *Config) GetCurrentContext(ctx context.Context) (string, *profile.Profile, types.NamespacedName, error) {
+
```
sttts, line 130: no empty line at beginning of a function

internal/config/config.go
```diff
@@ -122,6 +125,27 @@ func (c *Config) GetUpboundProfile(name string) (profile.Profile, error) {
...
+	if err != nil {
+		return "", nil, types.NamespacedName{}, err
+	}
+
+	profileName, curprof, ctp, err := profile.FromKubeconfig(ctx, profiles, conf)
+	if err != nil {
+		return "", nil, types.NamespacedName{}, err
+	}
+	return profileName, curprof, ctp, nil
+}
```
sttts, line 147:
```suggestion

	return profileName, curprof, ctp, nil
}
```

internal/config/config.go
Piotr1215: hmm, interesting why `golantci-lint` didn't complain :thinking:

Piotr1215: Looks like using `gofumpt` adds the required rules

internal/config/config.go
```diff
@@ -122,6 +125,26 @@ func (c *Config) GetUpboundProfile(name string) (profile.Profile, error) {
 	return p, nil
 }
 
+// GetCurrentContext returns the current context from the kubeconfig, profile and config
+func (c *Config) GetCurrentContext(ctx context.Context) (string, *profile.Profile, types.NamespacedName, error) {
```
sttts, line 129: add names to the return values. It's unclear what the string is.

sttts: also to FromKubeconfig please. I also forgot that.

Piotr1215: added return named values to both functions

cmd/up/controlplane/list.go
jeanduplessis: English translation: 
> This whole block will occur frequently. We should create a helper for this:

cmd/up/controlplane/list.go
```diff
@@ -19,82 +19,68 @@ import (
...
-			return err
-		}
-		ctpclient := cp.NewClient(cfg)
-		cfgclient := configurations.NewClient(cfg)
-
-		c.client = cloud.New(ctpclient, cfgclient, upCtx.Account)
-	}
-
+func (c *listCmd) AfterApply(kongCtx *kong.Context) error {
 	kongCtx.Bind(pterm.DefaultTable.WithWriter(kongCtx.Stdout).WithSeparator("   "))
```
RedbackThomson, line 40: Should we check for mutual exclusivity with `-g` and `-A`?

sttts: In kube `-A` trumps over `-n` afaik. No strong opinion whether we are stricter here.

RedbackThomson (approved):

PR #462, opened 2024-04-15
ctx: fix group+controlplane context code and add tests

sttts:
Signed-off-by: Dr. Stefan Schimanski <stefan.schimanski@upbound.com>

### DIFF
--- a/cmd/up/ctx/actions_test.go
+++ b/cmd/up/ctx/actions_test.go
@@ -1,0 +1,503 @@
+// Copyright 2024 Upbound Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ctx
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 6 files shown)


cmd/up/ctx/actions_test.go
```diff
@@ -0,0 +1,372 @@
...
+			},
+			wantLast: "upbound-previous",
+			wantErr:  "<nil>",
+		},
+	}
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			g := &Group{name: tt.group}
+			conf, last, err := g.accept(tt.conf, profile.Profile{KubeContext: "profile"}, tt.preferred)
+			if diff := cmp.Diff(tt.wantErr, fmt.Sprintf("%v", err)); diff != "" {
```
tnthornton, line 198: nit - instead of converting to a string, we could do something like 
```golang
if diff := cmp.Diff(tt.wantErr, err, test.EquateErrors())
```

cmd/up/ctx/actions_test.go
```diff
@@ -0,0 +1,372 @@
...
+			},
+			wantLast: "upbound-previous",
+			wantErr:  "<nil>",
+		},
+	}
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			ctp := &ControlPlane{NamespacedName: tt.ctp}
+			conf, last, err := ctp.accept(tt.conf, profile.Profile{KubeContext: "profile"}, "https://ingress", []byte{1, 2, 3}, tt.preferred)
+			if diff := cmp.Diff(tt.wantErr, fmt.Sprintf("%v", err)); diff != "" {
```
tnthornton, line 361: Similar comment for down here.

tnthornton (approved):
Thanks @sttts for all of the tests 🙇

I have a couple of nits, but nothing blocking.

cmd/up/ctx/actions_test.go
sttts: It's too strict. I know that Crossplane does it. But it checks the underlying type. IMO it is not what you want in most cases.

PR #471, opened 2024-04-26
Overhaul all commands to use stripped profiles and kubecontexts

RedbackThomson:

This pull request is the first step toward a larger paradigm shift in the way we use profiles and kube contexts to achieve a new, more unified user experience.

Customers will use the `up` CLI in two different ways:
- Setting their kubeconfig to point at a space, group or org:
  - `up ctx` to access the TUI, then navigating and pressing F10 to save
  - `up ctx -` to return to the previous context
  - `up ctx <organization>/<space>[/<group[/<name>]]` to set context at a specific path
- CRUD operations on resources within the scope of their newly set kubeconfig
  - `up ctp create/delete/list/get`
  - `up group create/delete/list/get`

Migrating to this new model allows for code paths to be greatly simplified, and the user experience to remain identical between self-hosted and cloud. Only `up ctx` will differ between the two environments, as cloud has higher levels of hierarchy (namely organizations and spaces). All commands run after setting context will be agnostic to the underlying host of the space (be it connected, managed or self-hosted).

This new mechanism eliminates the need for `up ctp kubeconfig get` and `up ctp connect`, replaced with the equivalent `up ctx` command. It also eliminates the need for most fields within the profiles, leaving only the token, account field and token type, which are unfortunately still needed to support commands like `up robot create` and `up configuration create`.

> Note: This PR does not produce a remotely functional or tested CLI, yet. Also I have not updated any of the tests, so all tests and the linting fails.

I have:


tnthornton:
>This new mechanism eliminates the need for up ctp kubeconfig get and up ctp connect

Still need to look at the PR, but how does one get a kubeconfig for say 'Argo' to point at a control plane with this paradigm?

RedbackThomson:
> > This new mechanism eliminates the need for up ctp kubeconfig get and up ctp connect
> 
> Still need to look at the PR, but how does one get a kubeconfig for say 'Argo' to point at a control plane with this paradigm?

`up ctx` does the same thing as `up ctp kubeconfig` (provided you have already set the session token in a profile):
```bash
up ctx --kubeconfig out.kubeconfig upbound-aws-us-east-1/my-group/my-ctp
```

cmd/up/controlplane/controlplane.go
```diff
@@ -52,6 +51,12 @@ func (c *Cmd) AfterApply(kongCtx *kong.Context) error {
 	}
 	kongCtx.Bind(upCtx)
 
+	client, err := upCtx.BuildCurrentContextClient()
+	if err != nil {
+		return errors.Wrap(err, "unable to get kube client")
+	}
+	kongCtx.Bind(client)
```
erhancagirici, line 58: kong is unable to bind interfaces via `Bind()` , instead use `BindTo()`
https://github.com/alecthomas/kong/issues/48#issuecomment-583695243

```suggestion
	kongCtx.BindTo(client, (*ctrlclient.Client)(nil))
```

cmd/up/controlplane/controlplane.go
```diff
@@ -117,23 +114,6 @@ local Spaces are supported. Use the "profile" management command to switch
...
-		resp.Updated,
-		resp.Synced,
-		resp.Ready,
-		resp.Message,
-		formatAge(resp.Age),
-	}
-}
-
 func extractSpaceFields(obj any) []string {
 	resp, ok := obj.(*controlplane.Response)
```
erhancagirici, line 118: We won't be operating on a `controlplane.Response` object anymore. This should use
```suggestion
	ctp, ok := obj.(spacesv1beta1.ControlPlane)
```
then extract fields from that struct.

cmd/up/group/cmd.go
```diff
@@ -52,10 +50,15 @@ func (c *Cmd) AfterApply(kongCtx *kong.Context) error {
...
-		// TODO: add legacy support
-		return errors.New("Only Spaces contexts supported for now.")
+	client, err := upCtx.BuildCurrentContextClient()
+	if err != nil {
+		return errors.Wrap(err, "unable to get kube client")
 	}
+
+	// todo(redbackthomson): Assert we are at the space level
+
+	kongCtx.Bind(client)
```
erhancagirici, line 60: kong is unable to bind interfaces via `Bind()` , instead use `BindTo()`
https://github.com/alecthomas/kong/issues/48#issuecomment-583695243

```suggestion
	kongCtx.BindTo(client, (*ctrlclient.Client)(nil))
```

erhancagirici (commented):
added a few quick findings that prevents running the commands.

cmd/up/controlplane/list.go
```diff
@@ -18,83 +18,39 @@ import (
...
-		p.Printfln("No Control planes found in %s group", c.deriveGroup())
-		return nil
-	}
-	if err != nil {
-		return err
+func (c *listCmd) Run(ctx context.Context, printer upterm.ObjectPrinter, p pterm.TextPrinter, upCtx *upbound.Context, client client.Client) error {
+	// todo(redbackthomson): Add support for `-A`
+
+	var l spacesv1beta1.ControlPlaneList
+	if err := client.List(ctx, &l); err != nil {
```
erhancagirici, line 46: In this current state of the code, the list operation already behaves like `-A` supplied, i.e. it is not respecting the group set in the current context.
Seems like the group should be passed as a `ListOption` to `client.List()`

cmd/up/controlplane/connect.go
```diff
@@ -1,109 +0,0 @@
...
-type connectCmd struct {
-	kubeconfig.ConnectionSecretCmd `cmd:""`
-}
-
-func (c *connectCmd) AfterApply(kongCtx *kong.Context, upCtx *upbound.Context) error {
-	return c.ConnectionSecretCmd.AfterApply(kongCtx, upCtx)
-}
-
-// Run executes the get command.
-func (c *connectCmd) Run(ctx context.Context, p pterm.TextPrinter, upCtx *upbound.Context, getter kubeconfig.ConnectionSecretGetter) error {
```
sttts, line 50: I wouldn't mind keeping a minimal command that just prints the alternative `up ctx` command line.

cmd/up/controlplane/controlplane.go
```diff
@@ -104,8 +103,6 @@ type Cmd struct {
 
 	PullSecret pullsecret.Cmd `cmd:"" help:"Manage package pull secrets."`
 
-	Kubeconfig kubeconfig.Cmd `cmd:"" name:"kubeconfig" help:"Manage control plane kubeconfig data."`
```
sttts, line 107: Same, keep it but print the alternative command. It's not obvious how it looks like, even for me it wasn't :D

cmd/up/ctx/navigation.go
```diff
@@ -60,41 +61,64 @@ func (f AcceptingFunc) Accept(ctx context.Context, upCtx *upbound.Context) error
...
-	ac := accounts.NewClient(cfg)
+	client := organizations.NewClient(cfg)
 
-	a, err := upbound.GetAccount(ctx, ac, upCtx.Account)
+	orgs, err := client.List(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "error listing organizations")
+	}
+
+	items := make([]list.Item, 0)
```
sttts, line 79:
```suggestion
	items := make([]list.Item, 0, len(orgs))
```

cmd/up/profile/profile.go
```diff
@@ -29,7 +29,6 @@ type Cmd struct {
 	Use     useCmd     `cmd:"" help:"Select an Upbound Profile as the default."`
 	View    viewCmd    `cmd:"" help:"View the Upbound Profile settings across profiles."`
 	Config  config.Cmd `cmd:"" help:"Interact with the current Upbound Profile's config."`
-	Set     setCmd     `cmd:"" help:"Set an Upbound Profile for use with a Space."`
```
sttts, line 32: keep command and output proper alternative (in this case there isn't any, so just tell the user).

cmd/up/trace/cmd.go
```diff
@@ -77,13 +77,16 @@ func (c *Cmd) AfterApply(kongCtx *kong.Context) error {
...
 	}
-	if c.Group == "" {
-		if !c.AllGroups {
-			c.Group = ns
+
+	ctp, exists := upCtx.ParseCurrentSpaceContextURL()
+
+	if c.Group == "" && !c.AllGroups {
+		if exists && ctp.Namespace != "" {
+			c.Group = ctp.Namespace
```
sttts, line 89: and if `!exists`?

internal/profile/profile.go
```diff
@@ -43,26 +38,18 @@ const (
...
 // A Profile is a set of credentials
 type Profile struct {
-	// ID is either a username, email, or token.
+	// ID is the referencable name of the profile.
 	ID string `json:"id,omitempty"`
 
-	// Type is the type of the profile.
-	Type Type `json:"type"`
+	// TokenType is the type of the profile.
+	TokenType TokenType `json:"type"`
```
sttts, line 45: does this mean an old up config.json will stop working? We should try to keep it usable.

cmd/up/ctx/actions.go
```diff
@@ -142,25 +136,16 @@ func (ctp *ControlPlane) Accept(ctx context.Context, upCtx *upbound.Context, pre
 
 		ingress = strings.TrimPrefix(groupCfg.Host, "https://")
 	} else {
-		// find existing space context
-		p, err := upCtx.Cfg.GetUpboundProfile(ctp.group.space.profile)
-		if err != nil {
-			return "", err
-		}
-		loader, err = p.GetSpaceKubeConfig()
```
ezgidemirel, line 150: we need to set the loader here, otherwise it panics when we try to get the raw config 
```
loader = upCtx.Kubecfg
```

cmd/up/controlplane/list.go
ezgidemirel:
```suggestion
	ns, _, err := upCtx.Kubecfg.Namespace()
	if err != nil {
		return errors.Wrap(err, "error getting namespace")
	}
	if err := cl.List(ctx, &l, &client.ListOptions{Namespace: ns}); err != nil {
		return errors.Wrap(err, "error getting control planes")
	}
```

cmd/up/controlplane/get.go
```diff
@@ -18,71 +18,39 @@ import (
...
-	ctp, err := c.client.Get(ctx, types.NamespacedName{Name: c.Name, Namespace: c.Group})
-	if controlplane.IsNotFound(err) {
-		p.Printfln("Control plane %s not found", c.Name)
-		return nil
-	}
-	if err != nil {
-		return err
+func (c *getCmd) Run(ctx context.Context, printer upterm.ObjectPrinter, p pterm.TextPrinter, upCtx *upbound.Context, client client.Client) error {
+	var ctp spacesv1beta1.ControlPlane
+	if err := client.Get(ctx, types.NamespacedName{Name: c.Name}, &ctp); err != nil {
```
ezgidemirel, line 47:
```suggestion
	ns, _, err := upCtx.Kubecfg.Namespace()
	if err != nil {
		return errors.Wrap(err, "error getting namespace")
	}

	if err := client.Get(ctx, types.NamespacedName{Name: c.Name, Namespace: ns}, &ctp); err != nil {
```

ezgidemirel:
Thanks a lot @RedbackThomson for putting together this PR. I've updated some parts in here [PR](https://github.com/RedbackThomson/up/pull/1) and tested some of the commands against a self-hosted space. 

https://github.com/upbound/up/assets/29144823/e1023a26-08ec-40b7-b18f-1e6303793943

https://github.com/upbound/up/assets/29144823/c7e0da21-cd2d-4505-a02d-d27096eeb8f1

https://github.com/upbound/up/assets/29144823/f35e64fb-3166-47f3-96bc-330fcbe71e4e

For `up ctx` command, I've noticed some differences with the previous implementation and discrepancies in context names while switching. I'm not sure if these are intended or not:

https://github.com/upbound/up/assets/29144823/4e405c8f-f19c-489f-a93e-694368ae1f88

cmd/up/controlplane/controlplane.go
RedbackThomson: Fixed in https://github.com/RedbackThomson/up/pull/1

cmd/up/controlplane/connect.go
RedbackThomson: What about a `--dry-run`? This would feel in line with kubectl

cmd/up/controlplane/controlplane.go
RedbackThomson: Same as above - `--dry-run`?

internal/profile/profile.go
RedbackThomson: I left the json tag the same. I just wanted to clarify in the go code that this type only applies to tokens now.

cmd/up/controlplane/connect.go
sttts: for which command?

erhancagirici:
Some findings from my side regarding `up ctx` at the current state of the code:
1. Not a bug but questionable UX: `up ctx` invocations always starts the navigation from org selection level, even when you are on some valid group level. I'd expect subsequent `up ctx`  invocations would start navigation from the current level I was currently in. Not sure if this intended, at least this was like that before.

2. When you select a controlplane via `up ctx` navigation, subsequent `up ctx` does not allow navigating anymore:
  - `Error: no matches for kind "ControlPlane" in version "spaces.upbound.io/v1beta1"` message is shown in interactive navigation
  - `up ctx ..` returns with no error and does not change the context.
  - `up ctx -` did not restore the previous context either, so basically I was stuck. I needed to manually restore context to the space cluster's kubeconfig via `kubectl config set-context`.
 - We need to fix navigation here.
 
3.  Commands should assert the relevant level and fail early for better UX. When I am on a controlplane context, `up ctp list` gives me `up: error: controlplane.listCmd.Run(): error getting control planes: no matches for kind "ControlPlane" in version "spaces.upbound.io/v1beta1"`   , I'd expect that this should inform me to navigate out of the controlplane level and do not try to issue the command.

cmd/up/controlplane/controlplane.go
sttts: Not sure I understand the proposal. What I mean is basically keep the commands in place, make them hidden, and replace `Run` with `Command X is not replaced by ....` output.

cmd/up/controlplane/connect.go
RedbackThomson: I am suggesting that to support printing the kubeconfig to stdout, we add a `--dry-run` flag to `up ctx`. That way all of the logic is the same, and the UX is the same, but instead of writing to file we simply print the new config.

sttts: dry-run is strange. Some ideas:
- `up ctx . -o yaml > kubeconfig`
- `up ctx . -f kubeconfig`

sttts: The original question is still open: we should keep these commands around, make them hidden, and print the alternative new command line cc @ezgidemirel @erhancagirici

cmd/up/controlplane/get.go
```diff
@@ -18,71 +18,48 @@ import (
...
-		return nil
-	}
-	if err != nil {
-		return err
+func (c *getCmd) Run(ctx context.Context, printer upterm.ObjectPrinter, p pterm.TextPrinter, upCtx *upbound.Context, client client.Client) error {
+	var ctp spacesv1beta1.ControlPlane
+	if err := client.Get(ctx, types.NamespacedName{Namespace: c.Group, Name: c.Name}, &ctp); err != nil {
+		if kerrors.IsNotFound(err) {
+			p.Printfln("Control plane %s not found", c.Name)
+			return nil
```
sttts, line 59: this is an error. It should lead to rc != 0.

cmd/up/controlplane/kubeconfig/connection.go
```diff
@@ -1,143 +0,0 @@
...
-const (
-	errFmtConfigBroken = "config is broken, missing %s: %q"
-)
-
-// ConnectionSecretCmd is the base for command getting connection secret for a control plane.
-type ConnectionSecretCmd struct {
-	Name  string `arg:"" required:"" help:"Name of control plane." predictor:"ctps"`
-	Token string `help:"API token used to authenticate. Required for Upbound Cloud; ignored otherwise."`
-	Group string `short:"g" help:"The control plane group that the control plane is contained in. By default, this is the group specified in the current profile."`
-}
```
sttts, line 44: wouldn't it be easy to keep this command? IMO it is still easier than some acrobatic with `up ctx`.

cmd/up/controlplane/kubeconfig/get.go
```diff
@@ -1,88 +0,0 @@
...
-	"github.com/upbound/up/internal/upbound"
-)
-
-// getCmd gets kubeconfig data for an Upbound control plane.
-type getCmd struct {
-	ConnectionSecretCmd
-
-	File    string `type:"path" short:"f" help:"File to merge control plane kubeconfig into or to create. By default it is merged into the user's default kubeconfig. Use '-' to print it to stdout.'"`
-	Context string `short:"c" help:"Context to use in the kubeconfig."`
-}
```
sttts, line 39: wouldn't it be easy to keep this command? IMO it is still easier than some acrobatic with up ctx.

cmd/up/controlplane/list.go
```diff
@@ -18,83 +18,49 @@ import (
...
-func (c *listCmd) Run(ctx context.Context, printer upterm.ObjectPrinter, p pterm.TextPrinter, upCtx *upbound.Context) error {
-	l, err := c.client.List(ctx, c.deriveGroup())
-	if controlplane.IsNotFound(err) {
-		p.Printfln("No Control planes found in %s group", c.deriveGroup())
-		return nil
-	}
-	if err != nil {
-		return err
+func (c *listCmd) Run(ctx context.Context, printer upterm.ObjectPrinter, p pterm.TextPrinter, upCtx *upbound.Context, cl client.Client) error {
+
```
sttts, line 54: nit: no empty lines at function start

cmd/up/controlplane/list.go
```diff
@@ -18,83 +18,49 @@ import (
...
+	var l spacesv1beta1.ControlPlaneList
+	if err := cl.List(ctx, &l, client.InNamespace(c.Group)); err != nil {
```
sttts, line 56: this will fail if inside a controlplane, right? We should create a good error.

cmd/up/controlplane/pkg/install.go
```diff
@@ -98,7 +98,7 @@ type installCmd struct {
 	Package string `arg:"" help:"Reference to the ${package_type}."`
 
 	// NOTE(hasheddan): kong automatically cleans paths tagged with existingfile.
-	Kubeconfig         string        `type:"existingfile" help:"Override default kubeconfig path."`
+	Kubeconfig         string        `hidden:"" type:"existingfile" help:"No longer used. Please use the KUBECONFIG environment variable instead."`
```
sttts, line 101: is it still working? If not, we should error.

cmd/up/ctx/cmd.go
```diff
@@ -330,3 +336,142 @@ func (c *Cmd) RunInteractive(ctx context.Context, kongCtx *kong.Context, upCtx *
 	}
 	return nil
 }
+
+func DeriveState(ctx context.Context, upCtx *upbound.Context, conf *clientcmdapi.Config) (NavigationState, error) {
```
sttts, line 340: if we don't have that yet, this deserves unit tests.

RedbackThomson: Oh yeah. A lot of this deserves unit tests. My #1 priority was just getting it working for now, but it needs some love (and bug finding)

cmd/up/ctx/cmd.go
```diff
@@ -330,3 +344,154 @@ func (c *Cmd) RunInteractive(ctx context.Context, kongCtx *kong.Context, upCtx *
...
+	default:
+		return &space, nil
+	}
+}
+
+// getOrgFromAuth loads the organization name by reading the JWT and pulling out
+// the claim. Returns an error if the JWT doesn't have the claim, or there was
+// an error parsing the JWT. This method DOES NOT verify the JWT against any
+// keys.
+func getOrgFromAuth(authToken string) (string, error) {
```
tnthornton, line 480: If we end up adding in Extensions to the kubeconfigs, this might be able to be handled there.

cmd/up/ctx/navigation.go
```diff
@@ -60,41 +67,43 @@ func (f AcceptingFunc) Accept(ctx context.Context, upCtx *upbound.Context) error
...
+	}
+
+	items := make([]list.Item, 0, len(orgs))
+	for _, org := range orgs {
+		items = append(items, item{text: org.DisplayName, kind: "organization", matchingTerms: []string{org.Name}, onEnter: func(ctx context.Context, upCtx *upbound.Context, m model) (model, error) {
+			m.state = &Organization{name: org.Name}
+			return m, nil
+		}})
 	}
+	sort.Sort(sortedItems(items))
```
tnthornton, line 92: Does this double sort?

cmd/up/group/cmd.go
```diff
@@ -15,18 +15,17 @@
 package group
 
 import (
-	"context"
 	"strconv"
 
 	"github.com/alecthomas/kong"
-	"github.com/crossplane/crossplane-runtime/pkg/errors"
+	"github.com/pkg/errors"
```
tnthornton, line 21: I think we still want `"github.com/crossplane/crossplane-runtime/pkg/errors"` here.

cmd/up/login/login.go
```diff
@@ -212,15 +212,15 @@ func setSession(ctx context.Context, p pterm.TextPrinter, upCtx *upbound.Context
 
 // constructAuth constructs the body of an Upbound Cloud authentication request
 // given the provided credentials.
-func constructAuth(username, token, password string) (*auth, profile.Type, error) {
+func constructAuth(username, token, password string) (*auth, profile.TokenType, error) {
```
tnthornton, line 215: Do we allow both token and password to be supplied? Seems like this will just result in permutations of `constructAuth("", token, "")` 🤔

cmd/up/organization/token.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/upbound/up-sdk-go/service/auth"
+	"github.com/upbound/up/internal/upbound"
+	"github.com/upbound/up/internal/upterm"
+)
+
+// tokenCmd generates an org-scoped token for use with spaces.
+type tokenCmd struct {
+	Upbound upbound.Flags `embed:""`
+
+	Name string `arg:"" env:"UPBOUND_ORGANIZATION" required:"" help:"Name of organization." predictor:"orgs"`
```
tnthornton, line 37: nit - do we need to prefix this env var? Are there other types of organizations?

internal/profile/profile.go
```diff
@@ -16,53 +16,38 @@ package profile
...
 )
 
 const (
 	NoSpacesContextMsg = "This cluster does not have spaces installed, use `up space init` to install spaces."
 	NoGroupMsg         = "The current kubeconfig context does not point to a group, use `up ctx` to select a group."
 )
 
-// Type is a type of Upbound profile.
-type Type string
+// TokenType is a type of Upbound session token format.
```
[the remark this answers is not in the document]
RedbackThomson: Yes these are just upbound api tokens - either user tokens or robot tokens.

cmd/up/ctx/cmd.go
RedbackThomson: That would be nice. This feels hacky and prone to breaking

cmd/up/space/init.go
```diff
@@ -280,32 +279,7 @@ func (c *initCmd) Run(ctx context.Context, upCtx *upbound.Context) error {
...
-	if err := upCtx.Cfg.AddOrUpdateUpboundProfile(upCtx.ProfileName, upCtx.Profile); err != nil {
-		return errors.Wrap(err, errUpdateProfile)
-	}
-	if err := upCtx.Cfg.SetDefaultUpboundProfile(upCtx.ProfileName); err != nil {
-		return errors.Wrap(err, errUpdateProfile)
-	}
-	if err := upCtx.CfgSrc.UpdateConfig(upCtx.Cfg); err != nil {
-		return errors.Wrap(err, errUpdateConfig)
-	}
+	// todo(redbackthomson): Update with `up ctx` equivalent code
```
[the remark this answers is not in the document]
RedbackThomson: I can probably follow up and add this as part of this PR. This feels critical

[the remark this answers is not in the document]
RedbackThomson: (I left these TODOs scattered everywhere to remind myself, more than anything)

cmd/up/ctx/navigation.go
RedbackThomson: I just copied the code 🤷🏼 I don't question it

cmd/up/space/init.go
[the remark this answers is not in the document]
RedbackThomson: Actually I think this is a no-op. For customers to use `up space init`, they are already pointed at a space hub cluster. We no longer care about adding anything to the profiles for spaces. They should be able to use `up ctx` straight away.

cmd/up/controlplane/connect.go
RedbackThomson: Got this feedback from @cwilhit too. I've put back all of the deprecated commands and replaced them with a stub simply returning an error and an appropriate redirection to their replacement (when possible).

RedbackThomson: Added support for `up ctx <> -f -` and `up ctx <> -f <file>`. Pretty easy, just needed to reorganise the code a bit.

cmd/up/controlplane/disconnect.go
```diff
@@ -15,88 +15,13 @@
...
-}
-
-func origContext(currentCtx string) (string, error) {
-	parts := strings.SplitN(currentCtx, "_", 4)
-	if len(parts) != 4 {
-		return "", fmt.Errorf(errFmtContextParts, len(parts))
-	}
-	return parts[len(parts)-1], nil
+func (c *disconnectCmd) Run(ctx context.Context) error {
+	return errors.New("this command has been removed in favor of 'up ctx'")
```
tnthornton, line 26: Similar question here as I had on connect.

internal/profile/profile.go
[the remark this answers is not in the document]
tnthornton: Right but session (user) tokens and robot tokens are disjoint. They're handled differently and even signed by a different trust chain.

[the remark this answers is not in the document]
RedbackThomson: I'm not exactly sure when you need to use robot tokens vs session tokens - maybe you could help fill me in, here?

If all of the existing CLI commands use session tokens, then I'd love to remove the need for the `type` field altogether. From the code, the only place I see the robot token being used is as part of `up login` which consequently sets the session of a profile. Not really sure how all of that works.

cmd/up/controlplane/connect.go
```diff
@@ -17,93 +17,15 @@ package controlplane
...
-	return nil
-}
-
-func ConnectControlplaneContextName(account string, name types.NamespacedName, origCtx string) string {
-	if name.Namespace == "" {
-		name.Namespace = "default" // passed by value. We can mutate it.
-	}
-	return fmt.Sprintf("%s%s_%s/%s_%s", upboundPrefix, account, name.Namespace, name.Name, origCtx)
+func (c *connectCmd) Run(ctx context.Context) error {
+	return errors.New("this command has been removed in favor of 'up ctx'")
```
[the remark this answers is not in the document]
RedbackThomson: Yeah ideally I'd tell them `run 'up ctx blah/blah/blah'`, but because we don't have the space name in the old commands it's not really possible to give the exact path.

cmd/up/controlplane/disconnect.go
RedbackThomson: Maybe at least here I could offer `run 'up ctx -'` to get them return to their previous context. That seems equivalent

internal/profile/profile.go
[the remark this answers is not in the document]
tnthornton: We technically can use either as the token in a kubeconfig to interact with the APIs. You're correct that `up login` gets you a session token, since it's for a user to login.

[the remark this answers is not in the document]
tnthornton: I think my preference would be to just remove confusion and remove the "session" piece of the comment since it could be either and robot != session token.

cmd/up/controlplane/connect.go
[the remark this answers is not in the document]
tnthornton: Could we do something like `up ctx <space name>/group/controlplaneName`? Let the end user fill in the blank for the space name.

cmd/up/ctx/actions_test.go
```diff
@@ -32,7 +35,7 @@ func TestGroupAccept(t *testing.T) {
 		wantLast  string
 		wantErr   string
 	}{
-		"ProfileToProfile": {
+		"ProfileToUpbound": {
```
RedbackThomson, line 38: Anything `*ToProfile` is no longer valid since this logic assumed we were pointing at the hub cluster, or could find the hub cluster information. Since this PR removes `profile.Kubeconfig` and `profile.Kubecontext`, we can't get back to the hub cluster. In the future we'll want to use something like Kubeconfig extensions to store hub cluster information for self-hosted spaces, but this will never be a valid use case for cloud.

cmd/up/ctx/actions_test.go
```diff
@@ -152,12 +76,12 @@ func TestGroupAccept(t *testing.T) {
 			wantConf: &clientcmdapi.Config{
 				CurrentContext: "upbound",
 				Contexts: map[string]*clientcmdapi.Context{
-					"upbound":          {Namespace: "other", Cluster: "profile", AuthInfo: "profile"},
-					"upbound-previous": {Namespace: "namespace1", Cluster: "upbound", AuthInfo: "upbound"},
+					"upbound":          {Namespace: "other", Cluster: "upbound", AuthInfo: "upbound"},
```
RedbackThomson, line 79: As of this PR, we no longer reference any existing clusters or auth info. Every time we accept, we will create a new cluster and authinfo. This makes the logic much simpler for cloud spaces since there's nothing to reference if you're starting with a fresh kubeconfig.

cmd/up/ctx/cmd.go
```diff
@@ -381,62 +390,67 @@ func DeriveState(ctx context.Context, upCtx *upbound.Context, conf *clientcmdapi
 		spaceKubeconfig = &config
 	}
 
+	if err := clientcmd.Validate(*spaceKubeconfig); err != nil {
+		return nil, err
+	}
```
RedbackThomson, line 395: This is a new requirement - the kubeconfig must be valid. This may be too strict, and we can revert it later if people have orphaned config in their kubeconfig.

internal/config/config_test.go
```diff
@@ -72,14 +72,6 @@ func TestAddOrUpdateUpboundProfile(t *testing.T) {
 				},
 			},
 		},
-		"Invalid": {
```
RedbackThomson, line 75: The only fields that were being validated have been removed. There is no permutation of `Profile` that is invalid, anymore.

RedbackThomson (commented):
Leaving some remarks about why certain tests have been modified

cmd/up/ctx/cmd_test.go
```diff
@@ -399,8 +399,8 @@ xPWjLExASVeAdNehjgFcrfoc7ZWtJYeE42his0athGjS/fNK7PnjijpZn6h76hRB
 			want: &Group{
 				Space: Space{
 					Name:    "hub",
-					Ingress: "https://hub", // TODO: there is nothing to store the hub (vs. the ingress)
-					CA:      []byte(hubCA),
+					Ingress: "https://ingress", // TODO: there is nothing to store the hub (vs. the ingress)
+					CA:      []byte(ingressCA),
```
RedbackThomson, line 403: With this PR, like I said above, we don't have the ability to get back to the hub cluster. So all of these tests will use the ingress and ingress CA.

cmd/up/ctx/cmd_test.go
```diff
@@ -302,3 +311,293 @@ func TestSwapContext(t *testing.T) {
...
+					"hub":     {Server: "https://hub", CertificateAuthorityData: []byte(hubCA)},
+					"ingress": {Server: "https://ingress", CertificateAuthorityData: []byte(ingressCA)},
+				},
+				AuthInfos: map[string]*clientcmdapi.AuthInfo{"hub": {Token: "token"}},
+			},
+			getIngressHost: func(ctx context.Context, cl client.Client) (host string, ca []byte, err error) {
+				return "https://ingress", []byte(ingressCA), nil
+			},
+			want: &Space{
+				Ingress: "https://ingress", // TODO: there is nothing to store the hub (vs. the ingress)
```
tnthornton, line 452: >// TODO: there is nothing to store the hub (vs. the ingress)

What is this meant to mean?

cmd/up/ctx/cmd_test.go
```diff
@@ -482,60 +481,31 @@ xPWjLExASVeAdNehjgFcrfoc7ZWtJYeE42his0athGjS/fNK7PnjijpZn6h76hRB
...
 						},
+						Name: "ctp1",
 					},
 					Name: "default",
 				},
 				Name: "ctp1",
 			},
 			wantErr: "<nil>",
 		},
-		"UpboundAPI": {
```
RedbackThomson: `up ctx` will never save upbound API or an organization scope into a kubeconfig. These situations shouldn't ever happen - we don't have any way to differentiate between these cases and a valid space.

tnthornton (approved):
I have one last question about a series of comments in one of the test files. If we could clarify that a bit, that would be awesome 👍

cmd/up/ctx/cmd_test.go
sttts: But we could differentiate by the URL, can't we?

cmd/up/ctx/cmd_test.go
```diff
@@ -482,60 +481,31 @@ xPWjLExASVeAdNehjgFcrfoc7ZWtJYeE42his0athGjS/fNK7PnjijpZn6h76hRB
...
 				Clusters: map[string]*clientcmdapi.Cluster{
-					"upbound": {Server: "https://eu-west-1.ibm-cloud.com", CertificateAuthorityData: []byte(hubCA)},
+					"upbound": {Server: "https://eu-west-1.ibm-cloud.com", CertificateAuthorityData: []byte(ingressCA)},
 				},
 				AuthInfos: map[string]*clientcmdapi.AuthInfo{"upbound": {Token: orgToken}},
 			},
-			getIngressHost: func(ctx context.Context, cl client.Client) (host string, ca []byte, err error) {
-				return "https://eu-west-1.ibm-cloud.com", []byte(ingressCA), nil
-			},
+			getIngressHost: ingressPublicNotFound,
```
sttts, line 501: this is wrong. The configmap will be visible very soon. Then up break.

cmd/up/ctx/cmd_test.go
```diff
@@ -482,60 +481,31 @@ xPWjLExASVeAdNehjgFcrfoc7ZWtJYeE42his0athGjS/fNK7PnjijpZn6h76hRB
...
 			want: &Group{
 				Space: Space{
 					Org: Organization{
 						Name: "org",
 					},
-					Name:     "space", // TODO: where does this come from?
-					Ingress:  "https://eu-west-1.ibm-cloud.com",
+					Name:     "eu-west-1", // TODO: where does this come from?
```
sttts, line 507: to answer the question: is that the hostname of the URL?

cmd/up/ctx/cmd.go
```diff
@@ -330,3 +349,183 @@ func (c *Cmd) RunInteractive(ctx context.Context, kongCtx *kong.Context, upCtx *
...
+
+	spaceClient, err := client.New(rest, client.Options{})
+	if err != nil {
+		return &Root{}, nil // nolint:nilerr
+	}
+
+	// determine if self-hosted by looking for ingress
+	host, ca, err := getIngressHost(ctx, spaceClient)
+	if kerrors.IsNotFound(err) || meta.IsNoMatchError(err) || kerrors.IsUnauthorized(err) {
```
sttts, line 409: this code happens to have broken disconnected Spaces.

PR #476, opened 2024-04-30
Prevent spinner from overwriting confirmation prompt during attach/detach

branden:

Attach and detach commands use a spinner to communicate the status of the ongoing operation. If a potential conflict is found, such as attaching a Space that's already been attached, the user is prompted to confirm whether they want to proceed. The spinner overwrites that prompt because it's still active, so we work around that by making the spinner draw its own prompt. But the first prompt is briefly visible for a flash, and the spinner's prompt may be misleading about whether the default is to proceed or abort.

This PR updates attach and detach commands to disable the spinner while prompting the user. This relieves the spinner of needing to write its own prompt, allowing attach/detach to use the existing prompt which correctly indicates the default selection.

I have:

I put together a [gist](https://gist.github.com/branden/f7fa37944d72ea86b9a024d86eb4fc8a) with output from the attach/detach commands. There are two text files `before.txt` and `after.txt` showing output before and after this PR. Each shows:
* successful attach
* reattach of attached cluster, aborted
* reattach of attached cluster, proceeding
* successful detach
* re-detach of detached cluster, aborted
* re-detach of detached cluster, proceeding

For convenience there's also a `diff.diff` showing the diff between `before.txt` and `after.txt`.

Note that the spinner's prompt in `before.txt` isn't captured due to competition between the spinner and the overwritten confirmation prompt and ensuing terminal weirdness.

### DIFF
--- a/cmd/up/space/attach.go
+++ b/cmd/up/space/attach.go
@@ -298,8 +298,7 @@ func (c *attachCmd) prepareSpace(ctx context.Context, attachSpinner *pterm.Spinn
 		}
 		ns, name := parts[0], parts[1]
 		if (space.Name != "" && space.Name != name) || space.Namespace != ns {
-			attachSpinner.UpdateText("Continue? (Y/n)")
-			if err := warnAndConfirm(
+			if err := warnAndConfirmWithSpinner(attachSpinner,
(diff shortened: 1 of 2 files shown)


darkmuggle (approved):
Pretty straightforward. 🚀

PR #491, opened 2024-05-06
Export context extensions for space metadata

RedbackThomson:

Create a new `SpaceExtension` type, defined as `runtime.Object`, to be used in `Context.Extensions` output by the `Accept` command. Stores a different configuration depending on whether the underlying space for the context is backed by Upbound cloud or is disconnected.

As of this PR, we don't use the extension data in any part of `up ctx` but in the future it will be used to:
- More easily derive the state (cloud vs. disconnected) without requiring lookups
- Point the `upbound` context at the space hub cluster, when using a disconnected space, instead of creating a new cluster using the ingress
- More easily derive the cloud organization without requiring [gross hacks](https://github.com/upbound/up/pull/483)

Example of a cloud kubeconfig with an extension:
```yaml
apiVersion: v1
clusters:
- cluster:
    insecure-skip-tls-verify: true
    server: https://upbound-gcp-us-west-1.space.mxe.upbound.services
  name: upbound
contexts:
- context:
    cluster: upbound
    extensions:
    - extension:
        apiVersion: spaces.upbound.io/v1alpha1
        kind: SpaceExtension
        spec:
          cloud:
            organization: upbound
      name: spaces.upbound.io/space
    namespace: nicks-group
    user: upbound
  name: upbound
current-context: upbound
kind: Config
preferences: {}
users:
- name: upbound
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1
      args:
      - organization
      - token
      command: up
      env:
      - name: ORGANIZATION
        value: upbound
      - name: UP_PROFILE
        value: staging
      interactiveMode: IfAvailable
      provideClusterInfo: false
```

I have:

Added some basic unit tests to assert the extensions are being added to the kubeconfig when accepting a group or ctp


cmd/up/ctx/navigation.go
```diff
@@ -338,48 +343,60 @@ func (ctp *ControlPlane) NamespacedName() types.NamespacedName {
...
 
+const (
+	ExtensionSpaceHub = "spaces.upbound.io/hub"
+	ExtensionSpace    = "spaces.upbound.io/space"
+)
+
 // buildSpacesClient creates a new kubeconfig hardcoded to match the provided
 // spaces access configuration and pointed directly at the resource.
-func buildSpacesClient(ingress string, ca []byte, authInfo *clientcmdapi.AuthInfo, resource types.NamespacedName) clientcmd.ClientConfig {
+func buildSpacesClient(space Space, resource types.NamespacedName) clientcmd.ClientConfig {
```
tnthornton, line 353: Do we need the resource parameter? Could we simply add the Namespace to the Space an operate just off of that?

tnthornton: I'm mostly trying to understand if we nearly have all of the information we need by passing the space or if there is expected to be other details from the resource that just don't make sense to include in the Space object.

tnthornton: Next question, do we need to pass the Space or could we make `buildSpacesClient` a method on the `Space` and derive the relevant information from that?

cmd/up/ctx/extension_types.go
```diff
@@ -0,0 +1,82 @@
...
+var (
+	// SpaceExtensionKind is kind of SpaceExtension
+	SpaceExtensionKind = reflect.TypeOf(SpaceExtension{}).Name()
+)
+
+func NewCloudV1alpha1SpaceExtension(org string) *SpaceExtension {
+	return &SpaceExtension{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       SpaceExtensionKind,
+			APIVersion: "spaces.upbound.io/v1alpha1",
```
tnthornton, line 60: nit - `upboundv1alpha1.SchemeGroupVersion.String()`

cmd/up/ctx/extension_types.go
```diff
@@ -0,0 +1,82 @@
...
+		},
+		Spec: &SpaceExtensionSpec{
+			Cloud: &CloudConfiguration{
+				Organization: org,
+			},
+		},
+	}
+}
+
+func NewDisconnectedV1alpha1SpaceExtension(hubCluster string) *SpaceExtension {
```
tnthornton, line 70: nit - 
```suggestion
func NewDisconnectedV1Alpha1SpaceExtension(hubCluster string) *SpaceExtension {
```

cmd/up/ctx/extension_types.go
```diff
@@ -0,0 +1,82 @@
...
+
+	Spec *SpaceExtensionSpec `json:"spec,omitempty"`
+}
+
+var (
+	// SpaceExtensionKind is kind of SpaceExtension
+	SpaceExtensionKind = reflect.TypeOf(SpaceExtension{}).Name()
+)
+
+func NewCloudV1alpha1SpaceExtension(org string) *SpaceExtension {
```
tnthornton, line 56: nit - 
```suggestion
func NewCloudV1Alpha1SpaceExtension(org string) *SpaceExtension {
```

cmd/up/ctx/navigation.go
```diff
@@ -338,48 +343,60 @@ func (ctp *ControlPlane) NamespacedName() types.NamespacedName {
 	return types.NamespacedName{Name: ctp.Name, Namespace: ctp.Group.Name}
 }
 
+const (
+	ExtensionSpaceHub = "spaces.upbound.io/hub"
+	ExtensionSpace    = "spaces.upbound.io/space"
```
sttts, line 348: doc or better name. What are these? The extension names on what? The context?

sttts:
> Point the upbound context at the space hub cluster, when using a disconnected space, instead of creating a new cluster using the ingress

Does this mean up will switch to the ingress as soon as I attach the Space, and it will not for disconnected? Or does this rather depend "where we are coming from", i.e. when on a hub context, we remember that fact via extension and come back to the hub later?

sttts:
Hmm, not working for me:
```
$ gcloud get credentials ....
$ up ctx
up: error: ctx.Cmd.Run(): failed to get API group resources: unable to retrieve the complete list of server APIs: v1: the server has asked for the client to provide credentials
```

cmd/up/ctx/navigation.go
RedbackThomson: Yeah I might move `buildSpacesClient` back to a struct method. I can't remember why I pulled it off but I've refactored it a bunch of times and landed back here.

FYI `resource` modifies the client depending on whether you specify a `resource.Name`:
- If you specify just a `namespace`, it points at the space hub and defines a namespace in the context
- If you specify a `name` as well, it points to a control plane and points to the `default` namespace in the context

I will update the GoDocs to clarify this.

RedbackThomson:
> Does this mean up will switch to the ingress as soon as I attach the Space, and it will not for disconnected? Or does this rather depend "where we are coming from", i.e. when on a hub context, we remember that fact via extension and come back to the hub later?

None of this is implemented as part of this PR, so I don't have a definitive answer for you yet. My idea was that we would always store a pointer to the hub cluster and whenever we point back to the group level, we would point at that cluster instead of creating a new one.

tnthornton (approved):

Issue #487, opened 2024-05-06
Fix `up ctx` for local and disconnected spaces


RedbackThomson:
### What happened?
@sttts ran the following commands and provided the output:

- `gcloud get credentials`
- `up ctx`  ✅
- `up ctx ./default` ✅ 
- `up ctx ./ctp-kine` ❌ 
- `up ctx .` broken due to upper command: `Kubeconfig context "upbound": Upbound organizations`, should be the group. ❌ 
- `up ctx` interactively going into ctp-kine ✅ 
- `up ctx .` broken again `Kubeconfig context "upbound": Upbound organizations`  ❌ 
- `up ctx -` has wrong output: `Kubeconfig context "upbound": Upbound organizations` , ❌  and I lost the group ❌ 
- `up ctx` – shows the organizations again although I am on the hub ❌ 

`up ctx` should work identically in cloud as it does in disconnected spaces.

RedbackThomson:
First part of the fix https://github.com/upbound/up/pull/491

RedbackThomson:
I can successfully validate that all of these commands are now working as expected! ❤️

PR #492, opened 2024-05-07
up space list against a disconnected space should return message to user

tnthornton:
Prior to this changeset, if a user was not logged in to Upbound Cloud they would be presented with the following error when pointed at a disconnected space:
```
up: error: space.listCmd.Run(): failed to get Account "upbound": Unauthorized: {"message":"Unauthorized"}
```

Changes:
- space list is for listing across spaces, which inherently doesn't work in isolation
- update current logic that returns an error if you are not logged in to upbound cloud
- add test coverage for the interactions
- introduce NopPrinters so that we can test logic without needing to account for stdout comms

Relates to https://github.com/upbound/up/issues/487

I have:

New unit tests.

With a local build:
1. Without logging in, attempted command and verified that I received expected message.
```
./_output/bin/darwin_arm64/up space list
You must be logged in and authorized to list Upbound Cloud Spaces
```
2. After logging in, verified that I could receive a list for the org that I was logged into.
```
./_output/bin/darwin_arm64/up alpha web-login
taylor logged in

./_output/bin/darwin_arm64/up space list
NAME                    MODE        PROVIDER   REGION
upbound-legacy-gcp      legacy      gcp        us-central-1
upbound-gcp-us-west-1   managed     gcp        us-west-1
upbound-aws-us-east-1   managed     aws        us-east-1
attached-gjsf9          connected
attached-mmdmz          connected
borrelli-self-hosted    connected
bulat-kind-space        connected
erhan-test-space        connected
ezgi-space              connected
michael-test            connected
milo-space-2            connected
milo-space-3            connected
sumbrizzle              connected

```
3. Logged out and attempted to list spaces again:
```
./_output/bin/darwin_arm64/up logout
taylor logged out

./_output/bin/darwin_arm64/up space list
You must be logged in and authorized to list Upbound Cloud Spaces
```
4. Verified that I could still install a space into a kind cluster.

### DIFF
--- a/cmd/up/main.go
+++ b/cmd/up/main.go
@@ -75,6 +75,7 @@ func (c *cli) AfterApply(ctx *kong.Context) error { //nolint:unparam
 	printer.Quiet = c.Quiet
 
 	ctx.Bind(printer)
+	ctx.BindTo(&printer, (*upterm.Printer)(nil))
 	ctx.Bind(c.Quiet)
 	return nil
 }
--- a/cmd/up/space/list.go
+++ b/cmd/up/space/list.go
@@ -35,7 +35,7 @@ import (
 var (
 	spacelistFieldNames = []string{"NAME", "MODE", "PROVIDER", "REGION"}
 
-	errListSpaces = "unable to list Upbound Cloud Spaces"
+	errListSpaces = "unable to list Upbound Spaces"
 )
 
 // listCmd lists all of the spaces in Upbound.
@@ -82,7 +82,7 @@ func (c *listCmd) Run(ctx context.Context, printer upterm.Printer, p pterm.TextP
 	if errors.As(err, &uerr) {
 		if uerr.Status == http.StatusUnauthorized {
 			p.Println("You must be logged in and authorized to list Upbound Cloud Spaces")
-			return nil
+			return uerr
 		}
 	}
 
--- a/cmd/up/space/list.go
+++ b/cmd/up/space/list.go
@@ -16,25 +16,34 @@ package space
 
 import (
 	"context"
+	"net/http"
 
 	"github.com/alecthomas/kong"
 	"github.com/pterm/pterm"
-	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/crossplane/crossplane-runtime/pkg/errors"
+
 	upboundv1alpha1 "github.com/upbound/up-sdk-go/apis/upbound/v1alpha1"
+	uerrors "github.com/upbound/up-sdk-go/errors"
 	"github.com/upbound/up-sdk-go/service/accounts"
+
 	"github.com/upbound/up/internal/upbound"
 	"github.com/upbound/up/internal/upterm"
 )
 
 var (
 	spacelistFieldNames = []string{"NAME", "MODE", "PROVIDER", "REGION"}
+
+	errListSpaces = "unable to list Upbound Cloud Spaces"
 )
 
 // listCmd lists all of the spaces in Upbound.
 type listCmd struct {
 	Upbound upbound.Flags `embed:""`
+
+	kc client.Client
+	ac *accounts.Client
 }
 
 // AfterApply sets default values in command after assignment and validation.
@@ -52,32 +61,39 @@ func (c *listCmd) AfterApply(kongCtx *kong.Context) error {
 	if err != nil {
 		return err
 	}
+	c.ac = accounts.NewClient(cfg)
+
+	kc, err := client.New(ctrlCfg, client.Options{})
+	if err != nil {
+		return errors.Wrap(err, errListSpaces)
+	}
+	c.kc = kc
 
 	kongCtx.Bind(upCtx)
-	kongCtx.Bind(ctrlCfg)
-	kongCtx.Bind(accounts.NewClient(cfg))
-
 	kongCtx.Bind(pterm.DefaultTable.WithWriter(kongCtx.Stdout).WithSeparator("   "))
 
 	return nil
 }
 
 // Run executes the list command.
-func (c *listCmd) Run(ctx context.Context, printer upterm.ObjectPrinter, p pterm.TextPrinter, upCtx *upbound.Context, ac *accounts.Client, rest *rest.Config) error {
-	a, err := upbound.GetOrganization(ctx, ac, upCtx.Account)
-	if err != nil {
-		return err
+func (c *listCmd) Run(ctx context.Context, printer upterm.Printer, p pterm.TextPrinter, upCtx *upbound.Context) error {
+	a, err := upbound.GetOrganization(ctx, c.ac, upCtx.Account)
+	var uerr *uerrors.Error
+	if errors.As(err, &uerr) {
+		if uerr.Status == http.StatusUnauthorized {
+			p.Println("You must be logged in and authorized to list Upbound Cloud Spaces")
+			return nil
+		}
 	}
 
-	sc, err := client.New(rest, client.Options{})
 	if err != nil {
-		return err
+		return errors.Wrap(err, errListSpaces)
 	}
 
 	var l upboundv1alpha1.SpaceList
-	err = sc.List(ctx, &l, &client.ListOptions{Namespace: a.Organization.Name})
+	err = c.kc.List(ctx, &l, &client.ListOptions{Namespace: a.Organization.Name})
 	if err != nil {
-		return err
+		return errors.Wrap(err, errListSpaces)
 	}
 
 	if len(l.Items) == 0 {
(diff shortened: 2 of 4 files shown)


cmd/up/space/list.go
```diff
@@ -52,32 +61,39 @@ func (c *listCmd) AfterApply(kongCtx *kong.Context) error {
...
 
 	return nil
 }
 
 // Run executes the list command.
-func (c *listCmd) Run(ctx context.Context, printer upterm.ObjectPrinter, p pterm.TextPrinter, upCtx *upbound.Context, ac *accounts.Client, rest *rest.Config) error {
-	a, err := upbound.GetOrganization(ctx, ac, upCtx.Account)
-	if err != nil {
-		return err
+func (c *listCmd) Run(ctx context.Context, printer upterm.Printer, p pterm.TextPrinter, upCtx *upbound.Context) error {
```
tnthornton, line 79: I really wish these parameters were a bit more composable. Saving that for a future clean up PR.

internal/upterm/printer.go
```diff
@@ -125,3 +132,46 @@ func (p *ObjectPrinter) printDefaultObj(obj any, fieldNames []string, extractFie
 	data[1] = extractFields(obj)
 	return p.TablePrinter.WithHasHeader().WithData(data).Render()
 }
+
+// NewNopObjectPrinter returns a Printer that does nothing.
+func NewNopObjectPrinter() Printer { return nopObjectPrinter{} }
```
tnthornton, line 137: This and the below NewNopTextPrinter were modeled after the NopLogger in crossplane-runtime.

cmd/up/main.go
```diff
@@ -75,6 +75,7 @@ func (c *cli) AfterApply(ctx *kong.Context) error { //nolint:unparam
 	printer.Quiet = c.Quiet
 
 	ctx.Bind(printer)
+	ctx.BindTo(&printer, (*upterm.Printer)(nil))
```
tnthornton, line 78: We need this for kong to respect ObjectPrinter as the upterm.Printer type.

cmd/up/space/list.go
```diff
@@ -52,32 +61,39 @@ func (c *listCmd) AfterApply(kongCtx *kong.Context) error {
...
+	c.ac = accounts.NewClient(cfg)
 
-	kongCtx.Bind(upCtx)
-	kongCtx.Bind(ctrlCfg)
-	kongCtx.Bind(accounts.NewClient(cfg))
+	kc, err := client.New(ctrlCfg, client.Options{})
+	if err != nil {
+		return errors.Wrap(err, errListSpaces)
+	}
+	c.kc = kc
```
tnthornton, line 70: Moved to composition instead of binding to make it easier to control what's happening in `Run` from the tests.

RedbackThomson (approved):

cmd/up/space/list.go
```diff
@@ -52,32 +61,39 @@ func (c *listCmd) AfterApply(kongCtx *kong.Context) error {
...
-	a, err := upbound.GetOrganization(ctx, ac, upCtx.Account)
-	if err != nil {
-		return err
+func (c *listCmd) Run(ctx context.Context, printer upterm.Printer, p pterm.TextPrinter, upCtx *upbound.Context) error {
+	a, err := upbound.GetOrganization(ctx, c.ac, upCtx.Account)
+	var uerr *uerrors.Error
+	if errors.As(err, &uerr) {
+		if uerr.Status == http.StatusUnauthorized {
+			p.Println("You must be logged in and authorized to list Upbound Cloud Spaces")
+			return nil
```
sttts, line 85: this is an error. Don't return nil, i.e. let the command return with rc != 0.

cmd/up/space/list.go
```diff
@@ -16,25 +16,34 @@ package space
...
 	"github.com/upbound/up-sdk-go/service/accounts"
+
 	"github.com/upbound/up/internal/upbound"
 	"github.com/upbound/up/internal/upterm"
 )
 
 var (
 	spacelistFieldNames = []string{"NAME", "MODE", "PROVIDER", "REGION"}
+
+	errListSpaces = "unable to list Upbound Cloud Spaces"
```
sttts, line 38: I bet @cwilhit would argue that there are not "Cloud Spaces" and there is no "Upbound Cloud", just "Upbound" and "Spaces".

internal/upterm/printer.go
```diff
@@ -26,6 +26,13 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+// Printer describes interactions for working with the ObjectPrinter below.
+// NOTE(tnthornton) ideally this would be called "ObjectPrinter".
+// TODO(tnthornton) rename this to ObjectPrinter.
```
sttts, line 31: what stops you from just naming it `ObjectPrinter`?

cmd/up/main.go
sttts: Why do we need both?

cmd/up/space/list.go
```diff
@@ -52,32 +61,39 @@ func (c *listCmd) AfterApply(kongCtx *kong.Context) error {
...
 	return nil
 }
 
 // Run executes the list command.
-func (c *listCmd) Run(ctx context.Context, printer upterm.ObjectPrinter, p pterm.TextPrinter, upCtx *upbound.Context, ac *accounts.Client, rest *rest.Config) error {
-	a, err := upbound.GetOrganization(ctx, ac, upCtx.Account)
-	if err != nil {
-		return err
+func (c *listCmd) Run(ctx context.Context, printer upterm.Printer, p pterm.TextPrinter, upCtx *upbound.Context) error {
+	a, err := upbound.GetOrganization(ctx, c.ac, upCtx.Account)
```
sttts, line 80: It surprises me that we have to do actual client calls to find out whether we can connect. Can't we just check for the existence of a profile? Every client call below can fail for whatever reason including not being authed. Having this extra check here seems superfluous.

cmd/up/space/list_test.go
```diff
@@ -0,0 +1,194 @@
...
+		},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+
+			err := tc.args.cmd.Run(
+				context.Background(),
+				upterm.NewNopObjectPrinter(),
+				upterm.NewNopTextPrinter(),
```
sttts, line 185: why do we need these extra nop constructs? Just print. Doesn't harm in the test, does it?

cmd/up/space/list.go
tnthornton: Agree - as long as we still get the nice human consumable message and not the standard one we've been seeing. I'll give that a try and report back.

cmd/up/space/list.go
tnthornton: Agree. So given that framing, I guess this would be `Upbound Spaces`.  My intent here was to distinguish to the user what they are attempting operate on. Easy enough to change the string depending on how folks are feeling 👍

internal/upterm/printer.go
tnthornton: A much larger changeset.

cmd/up/space/list.go
tnthornton: Good question. In theory this operation was already performed during login and we should have the org info in extensions. When that work lands, I think we can remove this.

cmd/up/space/list_test.go
tnthornton: Doesn't harm the test, but it does pollute the output of the test runs which makes it harder to see what has passed/failed.

cmd/up/main.go
tnthornton: If we don't Bind for ObjectPrinter type as well we get:
```
up: error: controlplane.listCmd.Run(): couldn't find binding of type upterm.ObjectPrinter for parameter 1 of func(context.Context, upterm.ObjectPrinter, pterm.TextPrinter, *upbound.Context, client.Client) error(), use kong.Bind(upterm.ObjectPrinter)
```

cmd/up/space/list.go
tnthornton: Ya I don't love this:
```
./_output/bin/darwin_arm64/up space list
You must be logged in and authorized to list Upbound Cloud Spaces

up: error: space.listCmd.Run(): failed to get Account "upbound": Unauthorized: {"message":"Unauthorized"}

```

But we can solve this in a different PR and at least have the error get handled appropriately.

cmd/up/space/list.go
RedbackThomson: > In theory this operation was already performed during login and we should have the org info in extensions

We wouldn't be referencing the kubeconfig for this command, since it doesn't access resources **inside** a space. This mechanism still uses profiles, hence the call to `GetOrganization`.

This goes back to our previous discussion about efficient ways to store information in profiles.

sttts: I actually misread. This is `GetOrganization` without an `s` and it is not just here for probing.

Issue #490, opened 2024-05-06
No warning when `up` isn't in the PATH


RedbackThomson:
### What happened?
When navigating using `up ctx` into a cloud space, and `up` is not present in the user's PATH, the user will see the following error:
```
up: error: unexpected argument tokenrces: unable to retrieve the complete list of server APIs: v1: Get "https://upbound-gcp-us-west-1.space.mxe.upbound.io/api/v1": getting credentials: exec: executable up failed with exit code 1
up: error: unexpected argument token
```

Ideally we would warn the user that `up` is not in their PATH since it practically makes `up` unusable.

### How can we reproduce it?
Remove `up` from PATH, run it separately, see error.

PR #495, opened 2024-05-07
Use `up` in exec if in PATH

RedbackThomson:

Fixes https://github.com/upbound/up/issues/490

The current logic for setting path to `up` in the Kubeconfig AuthInfo exec switches depending on the target release version. If we are on the `release` build, then use `"up"`. If we are on the `debug` build, then use the current executable path. This has unfortunate side effects when customers use `up` but don't have the CLI in `PATH` - the AuthInfo won't ever be able to resolve the exec.

This change standardises the mechanism regardless of release target. The AuthInfo will always fall back to the full path of the `up` executable, unless we can guarantee we're running it from `PATH`.

I have:

- [ ] ~~Added `backport release-x.y` labels to auto-backport this PR, as appropriate.~~

Running from `PATH`:
```yaml
$ up ctx acmeco/upbound-gcp-us-west-1/default --profile production -f -
...
users:
- name: upbound
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1
      args:
      - organization
      - token
      command: up
...
```

Running from pwd:
```yaml
$ ./up ctx upbound/upbound-gcp-us-west-1/nicks-group -f -
...
users:
- name: upbound
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1
      args:
      - organization
      - token
      command: /Users/nicholasthomson/Workspace/up/_output/bin/darwin_arm64/up
...
```

### DIFF
--- a/cmd/up/ctx/navigation.go
+++ b/cmd/up/ctx/navigation.go
@@ -17,6 +17,7 @@ package ctx
 import (
 	"context"
 	"os"
+	"os/exec"
 	"sort"
 
 	"github.com/charmbracelet/bubbles/list"
@@ -33,7 +34,6 @@ import (
 	"github.com/upbound/up-sdk-go/service/organizations"
 	"github.com/upbound/up/internal/profile"
 	"github.com/upbound/up/internal/upbound"
-	"github.com/upbound/up/internal/version"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


cmd/up/ctx/navigation.go
```diff
@@ -386,16 +386,16 @@ func buildSpacesClient(ingress string, ca []byte, authInfo *clientcmdapi.AuthInf
...
+	// find the current executable path
+	cmd, err := os.Executable()
+	if err != nil {
+		return nil, err
+	}
+
+	// if the current executable was the same `up` that is found in PATH
+	path, err := exec.LookPath("up")
+	if err == nil && path == cmd {
 		cmd = "up"
```
cbuto, line 398: what if path != cmd? do we need this check or should we always use whats returned by `os.Executable()`?

cbuto (approved):
this worked for me! 

small, non-blocking comment

cmd/up/ctx/navigation.go
RedbackThomson: `os.Exectuable` will always return the full absolute path to `up`. I think it's cleaner for it to just use `up` and rely on the PATH instead of hard coding it. That's how other CLIs do it, and I feel it's a good model.

cbuto: sounds good to me

Issue #485, opened 2024-05-06
`up ctx` fails if current-context is unset


RedbackThomson:
### What happened?
If the `current-context` set inside Kubeconfig is empty, `up ctx` will fail to derive the initial state. 

### How can we reproduce it?
`kubectl config unset current-context` then `up ctx`

PR #499, opened 2024-05-08
up ctx: allow activating context if current-context is empty

cbuto:

This PR handles the case where `up ctx upbound/upbound-gcp-us-west-1/default` is ran and there is no current-context (`k ctx -u`). Prior to these changes, the CLI would fail to swap the context and fail with:

```
up: error: ctx.Cmd.Run(): no "" context found
```

with these changes, the same command now runs successfully.

Fixes https://github.com/upbound/up/issues/485

I have:

updated unit tests + manual testing


cmd/up/ctx/cmd_test.go
```diff
@@ -202,6 +202,31 @@ func TestSwapContext(t *testing.T) {
...
+				},
+				Clusters:  map[string]*clientcmdapi.Cluster{"upbound": {Server: "server1"}, "upbound-previous": {Server: "server2"}, "other": {Server: "other"}},
+				AuthInfos: map[string]*clientcmdapi.AuthInfo{"upbound": {Token: "token1"}, "upbound-previous": {Token: "token2"}, "other": {Token: "other"}},
+			},
+			last:      "upbound-previous",
+			preferred: "upbound",
+			wantConf: &clientcmdapi.Config{
+				CurrentContext: "upbound",
+				Contexts: map[string]*clientcmdapi.Context{
+					"upbound": {Namespace: "namespace1", Cluster: "upbound", AuthInfo: "upbound"},
```
RedbackThomson, line 221: It's interesting that this deletes the `upbound-previous`. I'm not sure if this is what we would want? Idk...

One one hand, the "previous" was an unset current context so technically going back to that isn't supported. But on the other hand, if you retained `upbound-previous` you would basically have reconnected to the previous history.

Wdyt?

cbuto: yeah it wasn't immediately clear to me if this is expected, gonna see what happens when I try to go back to the last context after going through this flow

that logic is here:

https://github.com/upbound/up/blob/f90f11d11b1480274b50213652c848bc1290dfc2/cmd/up/ctx/cmd.go#L180-L189

RedbackThomson: Spoke with Casey on Slack:

We think that it makes sense to remove `upbound-previous` since moving from no current context to `upbound` represents the start of a new chain of history. It wouldn't make sense to return to the `upbound-previous` that was set before the empty current context. So this test assertion is expected

RedbackThomson (approved):

Issue #513, opened 2024-05-10
`up ctx` fails to grab a context when no kubecontext is set


cwilhit:
### What happened?

Follow Console instructions for grabbing the kubeconfig of a control plane, i.e.: 

`up ctx acmeco/upbound-gcp-us-west-1/default/ctp1`

It fails saying "up: error: ctx.Cmd.Run(): no "upbound" context found"

### How can we reproduce it?

```
kubectl config delete-context upbound
up ctx acmeco/upbound-gcp-us-west-1/default/ctp1
```

### What environment did it happen in?

Prod with `Client Version: v0.30.0-rc.0.16.gd077815`

RedbackThomson:
After deleting the `upbound` context, the `current-context` in the Kubeconfig was pointing at a non-existent context. When running, `up ctx` tried to resolve `current-context` and fell over because it couldn't be found. Ideally we should treat cases where the `current-context` doesn't resolve to anything the same way as we treat an empty `current-context`.

Issue #497, opened 2024-05-07
Name `up ctx` exported contexts with group and ctp


RedbackThomson:
### What happened?
Currently all contexts exported by `up ctx` use the name `upbound`. This context name provides no information about which space, group or control plane the context is pointed at without looking at the full kubeconfig. Ideally we should be putting some of that information into the context name when exporting it.

For example:
- A group should be `upbound/<group>`
- A ctp should be `upbound/<group>/<ctp>`

phisco:
Hit me multiple times already. AFAICT the standard is to use `_`, instead of `/`, other than that +1 on this! 🙏 
I gave it a try but unfortunately the code is not so straightforward and I got lost.

PR #518, opened 2024-05-10
Fix condition where current context points nowhere

RedbackThomson:

Fixes https://github.com/upbound/up/issues/513

Since we no longer depend on a valid current context, I thought the message that was output when a user switches contexts didn't make sense anymore. It was sometimes even showing `Kubeconfig context "" switched to: ...` which just feels like a mistake. So this PR also updates the output for a successful save and quit to be:
```
$ ./up ctx ./ctp1
Switched kubeconfig context to: Upbound upbound/upbound-aws-us-east-1/default/ctp1
```

The [plan](https://github.com/upbound/up/issues/497) is to eventually rename the exported context to include the group and ctp name, so we would show that instead of the breadcrumbs. But at least this describes that we're adding a new context.

I have:

### DIFF
--- a/cmd/up/ctx/cmd.go
+++ b/cmd/up/ctx/cmd.go
@@ -41,7 +41,7 @@ import (
 )
 
 const (
-	contextSwitchedFmt = "Added new context to kubeconfig: %s\n"
+	contextSwitchedFmt = "Switched kubeconfig context to: %s\n"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


cmd/up/ctx/cmd.go
```diff
@@ -41,7 +41,7 @@ import (
 )
 
 const (
-	contextSwitchedFmt = "Kubeconfig context %q switched to: %s\n"
+	contextSwitchedFmt = "Added new context to kubeconfig: %s\n"
```
cbuto, line 44: 🤔  we are doing more than added a context, we are changing the current context. 

what about:
```suggestion
	contextSwitchedFmt = "Switched kubeconfig context to %q\n"
```

RedbackThomson: Updated

branden (approved):

cbuto (approved):

PR #527, opened 2024-05-13
Stub out the configuration command and return an error

### LINKED ISSUE #493: `up cfg` has no usefulness in Spaces-based Upbound
### What happened?

`up cfg` is a set of commands to interact with Upbound Deployment Service and git-managed configurations. This feature doesn't exist in Cloud Spaces/Connected Spaces and there's no plan of record for making the feature available there currently. In the latest releases of `up`, because we're breaking with backwards compat for legacy Spaces, I suggest we leave the command stub there and print an error message explaining that it's not supported in Cloud/Connected Spaces.

adamwg:

The configuration command doesn't work with Upbound Cloud Spaces or Connected Spaces, and the CLI no longer works with Legacy Spaces, so remove the subcommands from `up configuration` and print a message.

Fixes #493

I have:

Ran the unit tests, and manually verified that `up configuration` prints the expected message.


cmd/up/main.go
```diff
@@ -105,7 +104,7 @@ type cli struct {
 	Space              space.Cmd                    `cmd:"" help:"Interact with Spaces."`
 	Group              group.Cmd                    `cmd:"" help:"Interact with groups inside Spaces."`
 	ControlPlane       controlplane.Cmd             `cmd:"" name:"controlplane" aliases:"ctp" help:"Interact with control planes in the current context, both in the cloud and in a local Space."`
-	Configuration      configuration.Cmd            `cmd:"" name:"configuration" aliases:"cfg" help:"Interact with configurations."`
+	Configuration      configuration.Cmd            `cmd:"" name:"configuration" aliases:"cfg" help:"Interact with configurations (deprecated)."`
```
RedbackThomson, line 107: I used `Deprecated: ` as a prefix for the help in `up ctp`.

https://github.com/upbound/up/blob/main/cmd/up/controlplane/controlplane.go#L116-L117

cmd/up/configuration/configuration.go
```diff
@@ -16,75 +16,15 @@ package configuration
...
-	Create   createCmd    `cmd:"" help:"Create a configuration."`
-	List     listCmd      `cmd:"" help:"List root configurations for the account."`
-	Get      getCmd       `cmd:"" help:"Get a single configuration for the account."`
-	Delete   deleteCmd    `cmd:"" help:"Delete a single configuration for the account."`
-	Template template.Cmd `cmd:"" help:"Interact with configuration templates."`
+}
 
-	Flags upbound.Flags `embed:""`
+func (c *Cmd) Run(ctx context.Context, p pterm.TextPrinter) error {
+	return errors.New("Configurations are not supported in Upbound Spaces")
```
RedbackThomson, line 29: nit: Just `Upbound` not `Upbound Spaces`. I can share a link to you about how we're changing the way we talk about the product, including naming schemes.

cmd/up/main.go
adamwg: :+1: Updatd.

cmd/up/configuration/configuration.go
adamwg: Thanks, makes sense. Feels a bit confusing to just say "not supported in Upbound", since it used to be supported, so I've updated it to `no longer supported in Upbound` to try and make it clearer why the command still exists. Open to suggestions.

RedbackThomson (approved):

PR #528, opened 2024-05-13
Update pterm so it doesn't intercept signals

### LINKED ISSUE #526: `up login` does not respond to Ctrl-C
### What happened?

When I run `up login`, I get a username prompt; if I don't want to enter a username and press Ctrl-C to kill the process, the `^C` literal shows up at the prompt:

```console
% up login
Username: ^C^C^C
```

The same happens at the `Password` prompt, but without the `^C` literal being echoed.

The only way to get out of `up login` is to either enter a username and password, or kill it externally.

### How can we reproduce it?

Run `up login`, then attempt to kill the command with `Ctrl-C` at the username  or password prompt.

### What environment did it happen in?

zsh in alacritty on Linux

adamwg:

In older versions, the pterm library we use for pretty-printing intercepted SIGINT and SIGTERM. This signal handling was set up in the pacakge's `init` function, so any code that ran before we set up our own signal handler couldn't be killed with Ctrl-C.

The signal handling was removed from pterm in a more recent version (see [1]). Updating to the latest fixes the issue where we couldn't kill `up login` with Ctrl-C (#526).

[1] https://github.com/pterm/pterm/issues/562

Fixes #526

Signed-off-by: Adam Wolfe Gordon <adam.wolfegordon@upbound.io>

I have:

* Manual testing to validate that `up login` can be killed with Ctrl-C at either the `Username` or the `Password` prompt, with or without entering input first.
* `make test` to ensure the dependency update didn't break any unit tests.
* Ran a few other commands (`up ctx`, `up ctp list`, `up space list`) to validate that output still looks right.

### DIFF
--- a/cmd/up/main.go
+++ b/cmd/up/main.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"io"
 	"os"
-	"os/signal"
 
 	"github.com/alecthomas/kong"
 	"github.com/pterm/pterm"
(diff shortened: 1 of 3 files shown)


tnthornton (approved):

adamwg:
Question: with this change, do we still need the signal handler we set up in `main.go`? All it does is call `os.Exit` (via `kongCtx.Exit`), and looking at git history I suspect it was introduced to work around the fact that pterm was intercepting our signals.

tnthornton:
>Question: with this change, do we still need the signal handler we set up in main.go? All it does is call os.Exit (via kongCtx.Exit), and looking at git history I suspect it was introduced to work around the fact that pterm was intercepting our signals.

My guess is no, but it should be easy enough to test your changes with the `up space init` flow to double check.

adamwg:
Confirmed that with our signal handler removed `up space init` can still be stopped with Ctrl-C. Added a commit to remove our signal handler, since it's not necessary.

PR #535, opened 2024-05-22
Get ingress host and CA from space through connect

RedbackThomson:

Closes https://github.com/upbound/team-orchestration/issues/99

Instead of connecting directly to the space using its FQDN and with skipping TLS, query the space through connect to get the ingress host and CA data. This allows for a TLS-secured connection to the space, and unlocks the ability to connect to "connected spaces" (assuming their host is accessible).

With the requirement to pass a context writer and now an ingress reader to a lot of different methods, I decided to wrap these into a new "context" object (`navCtx`) to make things cleaner - rather than passing many args into every method.

I have:

Blocked on testing in dev until https://github.com/upbound/spaces/pull/1067 is merged, but tested using stubbed responses.


avalanche123 (approved):

adamwg (approved):
lgtm. It would be nice to have some test coverage for the new code, but I'm not sure where we are with tests in `up` at this point (looks pretty limited).

internal/spaces/ingress.go
```diff
@@ -0,0 +1,120 @@
...
+		ingresses: make(map[types.NamespacedName]SpaceIngress),
+		bearer:    bearer,
+	}
+}
+
+func (c *ingressCache) Get(ctx context.Context, space v1alpha1.Space) (ingress *SpaceIngress, err error) {
+	nsn := types.NamespacedName{Name: space.Name, Namespace: space.Namespace}
+
+	// cache hit
+	if i, ok := c.ingresses[nsn]; ok {
```
cbuto, line 66: do we need a mutex here? or use something like https://github.com/hashicorp/golang-lru to handle it for us

RedbackThomson: I don't expect concurrent/multi-threaded access to any of these methods, but I am happy to move to using a purpose built cache just in case

cbuto (approved):

PR #537, opened 2024-05-23
spaces: Set and check the account name correctly

adamwg:

Previously, there were two issues preventing Spaces initialized via `up space init` (with its default arguments) from being attached to the console:

1. The account name could only be set by explicitly providing helm parameter flags (`--set` or `--file`). The `--account` flag and the account in the active profile were both ignored, resulting in spaces being created with the confusing default account name `notdemo`.
2. The check in `up space attach` was taking the current account name only from the `--account` parameter, rather than using the existing machinery that takes it from the active profile in the config file when necessary.

This PR introduces two fixes for these issues:

1. Update the `space init` code to fall back to the account in the upbound context object, which will be initialized appropriately based on `--account` and the config file. Let the helm parameters take precedence if they're given, since they're more specific.
2. Use the account name from the upbound context in `up space attach`, which should alwyas be populated appropriately.

While we're here, change the default account name to `disconnected`. This is at least a bit more meaningful than `notdemo`, since Spaces created without a valid account name can only be used as disconnected spaces (can't be attached to the console). Also warn users before using the default, since Spaces initialized this way can't be attached to the console and we assume most users will want that eventually.

I have:

Manually tested the `up space init` -> `up space attach` flow while logged in (profile set) and without a `.up/config.json` file (to trigger the new warning).


cmd/up/space/init.go
```diff
@@ -227,9 +225,20 @@ func (c *initCmd) AfterApply(kongCtx *kong.Context, quiet config.QuietFlag) erro
...
-func (c *initCmd) Run(ctx context.Context, upCtx *upbound.Context) error {
+func (c *initCmd) Run(ctx context.Context, upCtx *upbound.Context) error { //nolint:gocyclo
 	overrideRegistry(c.Registry.Repository.String(), c.helmParams)
-	ensureAccount(c.helmParams)
+	ensureAccount(upCtx, c.helmParams)
+
+	if c.helmParams["account"] == defaultAcct {
+		pterm.Warning.Println("No account name was provided. Spaces initialized without an account name cannot be attached to the Upbound console!")
+		confirm := pterm.DefaultInteractiveConfirm
+		confirm.DefaultText = "Would you like to proceed with the default account name?"
```
RedbackThomson, line 235: nit: Show that value in the prompt

cmd/up/space/init.go
```diff
@@ -227,9 +225,20 @@ func (c *initCmd) AfterApply(kongCtx *kong.Context, quiet config.QuietFlag) erro
...
 
 // Run executes the install command.
-func (c *initCmd) Run(ctx context.Context, upCtx *upbound.Context) error {
+func (c *initCmd) Run(ctx context.Context, upCtx *upbound.Context) error { //nolint:gocyclo
 	overrideRegistry(c.Registry.Repository.String(), c.helmParams)
-	ensureAccount(c.helmParams)
+	ensureAccount(upCtx, c.helmParams)
+
+	if c.helmParams["account"] == defaultAcct {
+		pterm.Warning.Println("No account name was provided. Spaces initialized without an account name cannot be attached to the Upbound console!")
```
RedbackThomson, line 233: Is it possible to change this value after the fact? Might be good to tell users "This cannot be changed after installation"

cmd/up/space/attach.go
```diff
@@ -150,7 +150,7 @@ func (c *attachCmd) Run(ctx context.Context, mgr *helm.Installer, kClient *kuber
 	} else if err != nil {
 		return errors.Wrapf(err, `failed to get ConfigMap "%s/%s"`, agentNs, mxpConfigMap)
 	}
-	if spaceAcc := mxpConfig.Data["account"]; spaceAcc != c.Upbound.Account {
+	if spaceAcc := mxpConfig.Data["account"]; spaceAcc != upCtx.Account {
```
sttts, line 153: :+1: – my fault

luxas (approved):
Thanks @adamwg!

PR #543, opened 2024-05-29
ctp: Allow version and upgrade channel to be specified in create

### LINKED ISSUE #541: create upbound saas controlplanes with specific version and/or upgrade channel
make it possible to create a controlplane with up cli with a specific uxp version and/or upgrade channel

Epic #

adamwg:

Add flags to specify the desired Crossplane version and upgrade channel in the
`up ctp create` command. Return an error if a version is specified with a
release channel other than `None`, since the server will accept a spec with a
version specified but ignore it in favor of the release channel's current
version.

For now, we don't have a programmatic way to fetch the valid release channels so
we hard-code the options. Similarly, we don't have an API to fetch supported
versions so we can't list them for the user in the CLI.

Fixes #541

I have:

Manual testing of the following cases:

* `up ctp create` without the new arguments works as before.
* `up ctp create --version=1.14.8-up.1 --channel=None` is successful and the ctp has the right version.
* `up ctp create --version=1.14.8-up.1` throws an error since the default release channel is `Stable`.
* `up ctp create --channel=Rapid` is successful and the ctp has the right release channel.

### DIFF
--- a/cmd/up/controlplane/create.go
+++ b/cmd/up/controlplane/create.go
@@ -16,8 +16,10 @@ package controlplane
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/alecthomas/kong"
+	"github.com/blang/semver/v4"
 	"github.com/pkg/errors"
 	"github.com/pterm/pterm"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -31,10 +33,34 @@ import (
 type createCmd struct {
 	Name  string `arg:"" required:"" help:"Name of control plane."`
 	Group string `short:"g" default:"" help:"The control plane group that the control plane is contained in. This defaults to the group specified in the current context"`
+
+	Crossplane struct {
+		Version     string `default:"" help:"The version of Universal Crossplane to use. The default depends on the selected auto-upgrade channel."`
+		AutoUpgrade struct {
+			Channel string `default:"Stable" help:"The Crossplane auto-upgrade channel to use. Must be one of: None, Patch, Stable, Rapid" enum:"None,Patch,Stable,Rapid"`
+		} `embed:""`
+	} `embed:"" prefix:"crossplane-"`
+
 	// todo(redbackthomson): Support all overrides for control planes
 	// ConfigurationName *string `help:"The optional name of the Configuration."`
 
 	SecretName string `help:"The name of the control plane's secret. Defaults to 'kubeconfig-{control plane name}'. Only applicable for Space control planes."`
+}
+
+// Validate performs custom argument validation for the create command.
+func (c *createCmd) Validate() error {
+	// TODO(adamwg): This validation should probably happen on the server side,
+	// at which point we could remove it here.
+	if c.Crossplane.Version != "" {
+		if c.Crossplane.AutoUpgrade.Channel != string(spacesv1beta1.CrossplaneUpgradeNone) {
+			return fmt.Errorf("upgrade channel must be %q to specify a version", string(spacesv1beta1.CrossplaneUpgradeNone))
+		}
+		_, err := semver.Parse(c.Crossplane.Version)
+		if err != nil {
+			return fmt.Errorf("invalid Crossplane version specified: %w; do not prefix the version with a 'v'", err)
+		}
+	}
+	return nil
 }
 
 // AfterApply sets default values in command after assignment and validation.
@@ -57,6 +83,16 @@ func (c *createCmd) Run(ctx context.Context, p pterm.TextPrinter, upCtx *upbound
 			Name:      c.Name,
 			Namespace: c.Group,
 		},
+	}
+
+	if c.Crossplane.Version != "" {
+		ctp.Spec.Crossplane.Version = &c.Crossplane.Version
+	}
+	if c.Crossplane.AutoUpgrade.Channel != "" {
+		ch := spacesv1beta1.CrossplaneUpgradeChannel(c.Crossplane.AutoUpgrade.Channel)
+		ctp.Spec.Crossplane.AutoUpgradeSpec = &spacesv1beta1.CrossplaneAutoUpgradeSpec{
+			Channel: &ch,
+		}
 	}
 
 	if err := client.Create(ctx, ctp); err != nil {
--- a/cmd/up/controlplane/create.go
+++ b/cmd/up/controlplane/create.go
@@ -41,9 +41,6 @@ type createCmd struct {
 		} `embed:""`
 	} `embed:"" prefix:"crossplane-"`
 
-	// todo(redbackthomson): Support all overrides for control planes
-	// ConfigurationName *string `help:"The optional name of the Configuration."`
-
 	SecretName string `help:"The name of the control plane's secret. Defaults to 'kubeconfig-{control plane name}'. Only applicable for Space control planes."`
 }
 


cmd/up/controlplane/create.go
```diff
@@ -31,12 +32,30 @@ import (
 type createCmd struct {
 	Name  string `arg:"" required:"" help:"Name of control plane."`
 	Group string `short:"g" default:"" help:"The control plane group that the control plane is contained in. This defaults to the group specified in the current context"`
+
+	Crossplane struct {
+		Version     string `default:"" help:"The version of Universal Crossplane to use. The default depends on the selected auto-upgrade channel."`
+		AutoUpgrade struct {
+			Channel string `default:"Stable" help:"The auto-upgrade channel to use. Must be one of: None, Patch, Stable, Rapid" enum:"None,Patch,Stable,Rapid"`
+		} `embed:""`
+	} `embed:""`
```
adamwg, line 41: Structured this to match the control plane API, so that it's easier to add additional options in the future without this struct becoming a huge, flat mess. Open to suggestions on the structure and naming of the flags - felt like `--crossplane-version` and `--crossplane-autoupgrade-channel` would be excessive, but I may have gone too far the other direction (`--channel` is a little cryptic?).

RedbackThomson: I like the code structure 👍🏼 

`channel` is not a property of the control plane, but of the crossplane instance installed into it. So my suggestion would be `--crossplane-channel` and `--crossplane-version`. Wdyt?

cmd/up/controlplane/create.go
```diff
@@ -31,12 +32,30 @@ import (
...
+
 	// todo(redbackthomson): Support all overrides for control planes
 	// ConfigurationName *string `help:"The optional name of the Configuration."`
```
RedbackThomson, line 44: We can probably delete this while we're at it

cmd/up/controlplane/create.go
```diff
@@ -31,12 +32,30 @@ import (
...
 	SecretName string `help:"The name of the control plane's secret. Defaults to 'kubeconfig-{control plane name}'. Only applicable for Space control planes."`
 }
 
+// Validate performs custom argument validation for the create command.
+func (c *createCmd) Validate() error {
+	// TODO(adamwg): This validation should probably happen on the server side,
+	// at which point we could remove it here.
+	if c.Crossplane.Version != "" && c.Crossplane.AutoUpgrade.Channel != string(spacesv1beta1.CrossplaneUpgradeNone) {
+		return fmt.Errorf("Upgrade channel must be %q to specify a version", string(spacesv1beta1.CrossplaneUpgradeNone))
+	}
```
RedbackThomson, line 55: Another validation that could happen here is checking to make sure the crossplane version does NOT have a leading `v`. `vX.Y.Z` is invalid - it should be just `X.Y.Z`. This is something we will eventually add to the server as well, but it is simple enough to add here. Otherwise trying to debug it, when applied, is a bit in the pain in the butt.

cmd/up/controlplane/create.go
```diff
@@ -59,6 +78,16 @@ func (c *createCmd) Run(ctx context.Context, p pterm.TextPrinter, upCtx *upbound
 		},
 	}
 
+	if c.Crossplane.Version != "" {
+		ctp.Spec.Crossplane.Version = &c.Crossplane.Version
```
RedbackThomson, line 82: Do you want to explicitly set the channel to none as well?

```
ctp.Spec.Crossplane.AutoUpgradeSpec = &spacesv1beta1.CrossplaneAutoUpgradeSpec{
	Channel: ptr.To(spacesv1beta1.CrossplaneUpgradeChannelNone),
}
```

cmd/up/controlplane/create.go
adamwg: I've been thinking of "a control plane" as being "a Crossplane", but from a product point of view I think that's not quite true (from a user point of view, there's a bit more to a control plane than just Crossplane). Adding the `crossplane-` prefix makes sense given that view of the world - updated to add it.

cmd/up/controlplane/create.go
adamwg: :+1: Makes sense to me. I've added a validation that parses the provided version into a semver, which will catch some errors beyond just the leading `v`. Assuming that will be the most common error, though, I've clarified it in the error message since `Invalid character(s) found` on its own could be a little confusing for users:

```
% up ctp create --crossplane-version=v1.14.8-up.1 --crossplane-channel=None ctp-with-invalid-version
up: error: controlplane (ctp) create: Invalid Crossplane version specified: Invalid character(s) found in major number "v1"; do not prefix the version with a 'v'
```

cmd/up/controlplane/create.go
adamwg: As it works now, to specify a version the user must also provide `--crossplane-channel=None`. I kind of lean toward keeping it explicit like that, since the default channel is `Stable` and a user might expect to be able to specify a version while also keeping the `Stable` channel. Not super opinionated on this point, though - we could implicitly change the upgrade channel when a version is provided without a channel being specified.

cmd/up/controlplane/create.go
```diff
@@ -36,12 +37,9 @@ type createCmd struct {
...
 		Version     string `default:"" help:"The version of Universal Crossplane to use. The default depends on the selected auto-upgrade channel."`
 		AutoUpgrade struct {
-			Channel string `default:"Stable" help:"The auto-upgrade channel to use. Must be one of: None, Patch, Stable, Rapid" enum:"None,Patch,Stable,Rapid"`
+			Channel string `default:"Stable" help:"The Crossplane auto-upgrade channel to use. Must be one of: None, Patch, Stable, Rapid" enum:"None,Patch,Stable,Rapid"`
 		} `embed:""`
-	} `embed:""`
-
-	// todo(redbackthomson): Support all overrides for control planes
-	// ConfigurationName *string `help:"The optional name of the Configuration."`
+	} `embed:"" prefix:"crossplane-"`
```
RedbackThomson, line 42: This is such a cool feature I didn't know about. TIL

cmd/up/controlplane/create.go
```diff
@@ -50,8 +48,14 @@ type createCmd struct {
 func (c *createCmd) Validate() error {
 	// TODO(adamwg): This validation should probably happen on the server side,
 	// at which point we could remove it here.
-	if c.Crossplane.Version != "" && c.Crossplane.AutoUpgrade.Channel != string(spacesv1beta1.CrossplaneUpgradeNone) {
-		return fmt.Errorf("Upgrade channel must be %q to specify a version", string(spacesv1beta1.CrossplaneUpgradeNone))
+	if c.Crossplane.Version != "" {
+		if c.Crossplane.AutoUpgrade.Channel != string(spacesv1beta1.CrossplaneUpgradeNone) {
+			return fmt.Errorf("Upgrade channel must be %q to specify a version", string(spacesv1beta1.CrossplaneUpgradeNone))
```
RedbackThomson, line 53: nit: All errors should begin with lowercase (the go error convention)

cmd/up/controlplane/create.go
```diff
@@ -50,8 +48,14 @@ type createCmd struct {
...
+		}
+		_, err := semver.Parse(c.Crossplane.Version)
+		if err != nil {
+			return fmt.Errorf("Invalid Crossplane version specified: %w; do not prefix the version with a 'v'", err)
```
RedbackThomson, line 57: here too

RedbackThomson (approved):

PR #546, opened 2024-05-31
ctx: Enable navigation between cloud and disconnected Spaces

### LINKED ISSUE #542: Navigate between cloud, connected, and disconnected spaces in `up ctx`

Currently, the selection of contexts available in `up ctx` depends on the currently selected kubeconfig context:

* If there's no kubeconfig, `up ctx` shows available cloud spaces.
* If the current kubeconfig context doesn't have a space installed in it (i.e., is a normal Kubernetes cluster without Spaces installed), `up ctx` shows available cloud spaces.
* If the current kubeconfig context is a cloud space, `up ctx` shows available cloud spaces.
* If the current kubeconfig context is a self-hosted space, `up ctx` shows only groups and control planes in that space (regardless of whether the space is connected or disconnected). The only way to get back to the list of cloud spaces is to manually change the kubeconfig context outside of `up`.

Make it possible to navigate between cloud spaces, connected spaces, and self-hosted spaces regardless of the current kubeconfig context.

adamwg:
In the root/home screen of `up ctx`, include "Disconnected Spaces" alongside the list of cloud organizations. The "Disconnected Spaces" section lists all kubeconfig contexts that contain a Space, as determined by either the spaces extension being present in the config, or the mxp ConfigMap being present in the cluster.
  
Since the home screen is now available regardless of the user's current kubeconfig context, display a message (rather than an error) if we're unable to list organizations from cloud. This ensures that users who aren't logged in (because they use only disconnected Spaces) can still use `up ctx`.

This also makes it necessary to improve how we handle cursor manipulation in the list view. We programmatically manipulate the cursor in two cases:
  
1. When constructing the list of items, to place the cursor below the "back" item in the list if it exists.
2. When the cursor comes across an unselectable item, to "skip" it.
  
We now need to also handle unselectable items in case (1), and case (2) didn't correctly handle lists with unselectable items at the beginning or end.
  
Consolidate this handling by introducing some helper functions that handle the cursor movement and correctly handle the corner cases. We handle unselectable items at the beginning or end of lists by "bouncing" back in the direction the user was moving until we hit a selectable item; in the corner case where all items are unselectable we'll end up with an unselectable item selected, which is probably the best we can do given the constraints of the list library.

Fixes #542

### DIFF
--- a/cmd/up/ctx/list.go
+++ b/cmd/up/ctx/list.go
@@ -208,9 +208,7 @@ func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) { // nolint:gocyclo // T
 			switch {
 			case key.Matches(msg, backNavBinding):
 				if state, ok := m.state.(Back); ok {
-					if state.CanBack() {
-						fn = state.Back
-					}
+					fn = state.Back
 				}
 			case key.Matches(msg, selectNavBinding):
 				if i, ok := m.list.SelectedItem().(item); ok {
--- a/cmd/up/ctx/list.go
+++ b/cmd/up/ctx/list.go
@@ -137,15 +137,14 @@ func NewList(items []list.Item) list.Model {
 	l.KeyMap.ShowFullHelp = key.NewBinding(key.WithDisabled())
 	l.KeyMap.CloseFullHelp = key.NewBinding(key.WithDisabled())
 
-	// check for initial cursor conditions
-	if len(items) > 1 {
-		nested := items[0].(item).back
-
-		if nested {
-			// move the cursor down below the '..' button
-			l.CursorDown()
+	// Try not to start on a back button or unselectable item.
+	moveCursorDownUntil(&l, func(cur list.Item) bool {
+		i, ok := cur.(item)
+		if !ok {
+			return true
 		}
-	}
+		return !i.back && !i.notSelectable
+	})
 
 	return l
 }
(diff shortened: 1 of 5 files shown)


adamwg:
Opened this by accident when testing a feature in the `gh` CLI, so marking as draft while I clean it up to be ready for review :smile:

cmd/up/ctx/list.go
```diff
@@ -55,20 +55,23 @@ var quitBinding = key.NewBinding(
 
 type KeyFunc func(m model) (model, error)
 
+type itemPadding struct {
```
RedbackThomson, line 58: nit: Could just make this a generic `padding` and it becomes useful in other places

go.mod
```diff
@@ -46,6 +46,7 @@ require (
 	google.golang.org/api v0.164.0
 	gopkg.in/yaml.v2 v2.4.0
 	gopkg.in/yaml.v3 v3.0.1
+	gotest.tools/v3 v3.5.1
```
adamwg, line 49: I didn't see an assertion library in use yet, so I picked my favorite. I somewhat prefer using one vs. raw `t.Fatal`s, but don't feel super strongly about which one to use, so let me know if there's an Upbound-wide preference.

cmd/up/ctx/list.go
```diff
@@ -272,19 +260,59 @@ func (m model) moveToSelectableItem(msg tea.Msg) list.Model {
...
+
+	return m.list
+}
+
+// moveCursorDownUntil moves the model's cursor down until cond returns true for
+// the selected item. If the bottom of the list is reached without cond
+// returning true, the cursor will be moved back up to the last item for which
+// cond is true. If no such items exist the first item in the list will be
+// selected.
+func moveCursorDownUntil(m *list.Model, cond func(list.Item) bool) {
```
RedbackThomson, line 283: I think we're re-using the inner anonymous type enough that it could justify its own type definition. eg.

```go
type moveCursorConditionFn func(list.Item) bool

func moveCursorDownUntil(m *list.Model, cond moveCursorConditionFn) {
  ...
}
```

cmd/up/ctx/list.go
```diff
@@ -272,19 +260,59 @@ func (m model) moveToSelectableItem(msg tea.Msg) list.Model {
 		return m.list
 	}
 
+	switch {
+	case key.Matches(keyMsg, m.list.KeyMap.CursorUp):
+		moveCursorUpUntil(&m.list, func(cur list.Item) bool {
+			return !cur.(item).notSelectable
+		})
```
RedbackThomson, line 267: nit: Could make this inner anonymous function into a static function so this becomes a bit clearer. I'm envisioning something like:

```go
moveCursorUpUntil(&m.list, itemSelectable)
```

cmd/up/ctx/list_test.go
```diff
@@ -0,0 +1,249 @@
...
+
+import (
+	"testing"
+
+	"github.com/charmbracelet/bubbles/list"
+	tea "github.com/charmbracelet/bubbletea"
+	"gotest.tools/v3/assert"
+)
+
+func TestNewList_CursorPlacement(t *testing.T) {
```
RedbackThomson, line 25:
```suggestion
func TestNewList_InitialCursorPlacement(t *testing.T) {
```

cmd/up/ctx/list_test.go
```diff
@@ -0,0 +1,249 @@
...
+			expectedIndex: 0,
+		},
+		"selectable item, down": {
+			items: []list.Item{
+				item{text: "item 1"},
+				item{text: "item 2"},
+			},
+			startingIndex: 0,
+			key:           'j',
+			expectedIndex: 0,
```
RedbackThomson, line 134: Shouldn't this be `1`?

cmd/up/ctx/list_test.go
```diff
@@ -0,0 +1,249 @@
...
+		},
+		"selectable item, up": {
+			items: []list.Item{
+				item{text: "item 1"},
+				item{text: "item 2"},
+			},
+			startingIndex: 1,
+			key:           'k',
+			expectedIndex: 1,
```
RedbackThomson, line 143: Shouldn't this be `0`?

cmd/up/ctx/list_test.go
RedbackThomson: Am I misunderstanding these cases? We are starting at index 0 and pressing j to go down - we'd get to index 1?

cmd/up/ctx/navigation.go
```diff
@@ -131,6 +150,115 @@ func (r *Root) Breadcrumbs() string {
...
+
+func spaceItemFromKubeContext(ctx context.Context, kubeconfig clientcmdapi.Config, name string) (item, bool) {
+	kubectx, ok := kubeconfig.Contexts[name]
+	if !ok {
+		return item{}, false
+	}
+
+	// A kubeconfig context represents a disconnected space in two cases:
+	// 1. The context has the spaces extension, with the Disconnected field
+	//    filled in.
```
RedbackThomson, line 191: AFAIK this would only ever apply to the `upbound` and `upbound-previous`, since we never write to any other contexts. But we wouldn't want to show either of these contexts in this list - they will already be accounted for by the other contexts in the file.

cmd/up/ctx/navigation.go
```diff
@@ -131,6 +150,115 @@ func (r *Root) Breadcrumbs() string {
...
+
+	spacesExt, err := upbound.GetSpaceExtension(kubectx)
+	if err != nil {
+		return item{}, false
+	}
+
+	var hubContextName string
+	if spacesExt == nil {
+		hubContextName = name
+	} else if spacesExt.Spec.Disconnected != nil {
```
RedbackThomson, line 206: Possible NPE if `Spec == nil`

cmd/up/ctx/list_test.go
adamwg: These are a bit unintuitive. `moveToSelectableItem` gets called *after* the TUI framework has handled the key and moved the cursor. In the case where it's moved to a selectable item, we don't need to do anything - it's just when we land on something unselectable that we do further movement.

cmd/up/ctx/list.go
adamwg: Good call - done.

cmd/up/ctx/navigation.go
adamwg: Aha, I hadn't noticed that, but looking at how the extension is used I see it's true. Given this, we can ignore any context with the extension, and we don't have to worry about duplicate context names (the deduplication was there to handle cases where `upbound` and/or `upbound-previous` pointed to a disconnected Space). Both of those things simplify the logic somewhat, which is nice.

go.mod
sttts: was advocating for `require` elsewhere. There is some style guide somewhere (Crossplane?) that says to avoid a library. But we don't do oursevles a favor with that. Tests are so much more readible by some assertion library.

PR #552, opened 2024-06-05
ctx: Improve performance and make UI more responsive

### LINKED ISSUE #549: Navigating to Disconnected Spaces in `up ctx` can be slow/hang with many or unreachable kubeconfig contexts
### What happened?

`up ctx` lists disconnected spaces by (serially) connecting to each context in the user's kubeconfig and attempting to get the configmap that stores ingress details for Spaces. If you have a lot of contexts, or if some of them are slow/unreachable, this can by very slow or hang entirely.

### How can we reproduce it?

Run `up ctx` and navigate to `Disconnected Spaces` when your kubeconfig contains a context whose cluster is unreachable from your network.

### Proposed Fixes

* Handle some number (say 100) contexts in parallel.
* Time-bound each request and the whole loop so we don't hang for too long.
* Show an error if we fail/timeout probing some context, so the user isn't surprised if a space "disappears" from the list (present one day, gone the next).
* Possibly add some UI indication that work is being done (a spinner?) so it doesn't look frozen if it's taking 1-2 seconds.

adamwg:

Use a timeout of 2 seconds when trying to fetch the ingress information for a disconnected Space and attempt to fetch them all in parallel. This should prevent us from hanging forever if a particular kubeconfig context is unreachable or misbehaving. To further help with slow contexts, avoid the overhead of controller-runtime's REST mapping by using the typed client-go client to fetch the ingress ConfigMap.

To make the UI feel more responsive when requests are slow, update the item list asynchronously to the main update loop of the TUI and trigger an update once we have our items. This lets us run a spinner while the items are being updated. As part of this change, switch from generating our own title to using the list widget's built-in title bar, which doesn't change how the TUI looks at all, but lets us leverage the built-in spinner capability of the list widget.

Fixes #549

I have:

To test the handling of slow or broken contexts, I did the following:

1. Added 50 contexts to my kubeconfig pointing at `localhost:65432`, which doesn't have any cluster listening on it.
2. Ran `nc -kl 65432` to simulate a server accepting requests but not responding.

This lets me observe the new timeout, which I verified we're hitting by adding some temporary debug logging.

Also tested navigating through the tree and selecting contexts as normal to ensure the spinner change didn't break anything.

### DIFF
--- a/cmd/up/ctx/navigation.go
+++ b/cmd/up/ctx/navigation.go
@@ -30,6 +30,7 @@ import (
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 	"k8s.io/utils/ptr"
@@ -210,7 +211,7 @@ func spaceItemFromKubeContext(ctx context.Context, kubeconfig clientcmdapi.Confi
 		return nil, err
 	}
 
-	cl, err := client.New(rest, client.Options{})
+	cl, err := corev1client.NewForConfig(rest)
 	if err != nil {
 		return nil, err
 	}
--- a/cmd/up/ctx/navigation.go
+++ b/cmd/up/ctx/navigation.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"os/exec"
 	"sort"
+	"sync"
 	"time"
 
 	"github.com/charmbracelet/bubbles/list"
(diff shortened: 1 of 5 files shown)


cmd/up/ctx/navigation.go
```diff
@@ -160,53 +164,79 @@ func (d *Disconnected) Items(ctx context.Context, upCtx *upbound.Context, navCtx
...
-			continue
-		}
-		if spacesExt != nil {
-			// This is an up-managed context, which means it's either a cloud
-			// Space, or a disconnected Space represented by some other
-			// kubeconfig context, which we'll find later.
-			continue
-		}
+	var wg sync.WaitGroup
+	var mu sync.Mutex
```
adamwg, line 168: Thought about using a channel here to pass results back from the goroutines, but it felt simpler (and probably equally performant) to just lock.

internal/profile/kubeconfig.go
```diff
@@ -18,18 +18,18 @@ import (
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 )
 
 // GetIngressHost returns the ingress host of the Spaces cfg points to. If the
 // ingress is not configured, it returns an empty string.
-func GetIngressHost(ctx context.Context, cl client.Client) (host string, ca []byte, err error) {
-	mxpConfig := &corev1.ConfigMap{}
-	if err := cl.Get(ctx, types.NamespacedName{Name: "ingress-public", Namespace: "upbound-system"}, mxpConfig); err != nil {
+func GetIngressHost(ctx context.Context, cl corev1client.ConfigMapsGetter) (host string, ca []byte, err error) {
```
RedbackThomson, line 27: Any particular reason why we changed from client to typed clientset?

RedbackThomson (approved):

internal/profile/kubeconfig.go
adamwg: This was a suggestion from @sttts to reduce the number of roundtrips the client needs to make, since the controller-runtime client does some discovery calls before it actually fetches the ConfigMap.

PR #554, opened 2024-06-05
ctx: Look up cloud/connected Space ingresses in parallel

adamwg:

Similar to our recent change for disconnected Spaces, make the organization view more responsive by looking up all the cloud and connected Space ingresses in parallel.

I have:

Manual testing with the two orgs I'm part of.

### DIFF
--- a/cmd/up/ctx/navigation.go
+++ b/cmd/up/ctx/navigation.go
@@ -16,7 +16,6 @@ package ctx
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -313,10 +312,9 @@ func (o *Organization) Items(ctx context.Context, upCtx *upbound.Context, navCtx
 
 				ingress, err := navCtx.ingressReader.Get(ctx, space)
 				if err != nil {
-					if errors.Is(err, spaces.SpaceConnectionError) {
-						// we found the space to be unreachable
-						continue
-					}
+					// TODO(adamwg): Add an unselectable item for the Space with
+					// relevant text depending on the type of error.
+					continue
 				}
 
 				mu.Lock()
--- a/cmd/up/ctx/navigation.go
+++ b/cmd/up/ctx/navigation.go
@@ -287,6 +287,9 @@ func (o *Organization) Items(ctx context.Context, upCtx *upbound.Context, navCtx
 
 	items := make([]list.Item, 0)
 	items = append(items, item{text: "..", kind: o.BackLabel(), onEnter: o.Back, back: true})
+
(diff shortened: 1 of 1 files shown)


cmd/up/ctx/navigation.go
```diff
@@ -300,24 +303,33 @@ func (o *Organization) Items(ctx context.Context, upCtx *upbound.Context, navCtx
...
+		wg.Add(1)
+		go func(space upboundv1alpha1.Space) {
+			defer wg.Done()
+			ingress, err := navCtx.ingressReader.Get(ctx, space)
+			if err != nil {
+				if errors.Is(err, spaces.SpaceConnectionError) {
+					// we found the space to be unreachable
+					return
+				}
 			}
```
adamwg, line 315: @RedbackThomson Unchanged in this PR, but is it intentional that we proceed to add an item if we get an error other than `SpaceConnectionError`? I suspect we might NPE below (`*ingress`) in some cases.

cmd/up/ctx/navigation.go
```diff
@@ -300,24 +303,33 @@ func (o *Organization) Items(ctx context.Context, upCtx *upbound.Context, navCtx
 			continue
 		}
 
-		ingress, err := navCtx.ingressReader.Get(ctx, space)
-		if err != nil {
-			if errors.Is(err, spaces.SpaceConnectionError) {
-				// we found the space to be unreachable
-				continue
+		wg.Add(1)
```
RedbackThomson, line 306: Can we cap the wait group capacity to prevent this from spooling up too much if an org has a lot of connected spaces?

RedbackThomson (approved):

cmd/up/ctx/navigation.go
RedbackThomson: Hmmm I can't think of a reason why we would want to continue, so this would be an oversight. Ideally we wouldn't want to fail silently, though, since users just wouldn't see the space but also wouldn't know why

cmd/up/ctx/navigation.go
adamwg: Yeah, makes sense. Capped at 20 for now, chosen arbitrarily but with the thinking that it will let us fetch everything in parallel for the majority of users while keeping the number of concurrent requests reasonable.

cmd/up/ctx/navigation.go
adamwg: Made the `continue` unconditional for now, and left a TODO to add unselectable items for the unreachable spaces. I think we want to show them regardless of what kind of error we hit, but may want different wording depending on whether it's a known error (`SpaceConnectionError`) or something else.

PR #555, opened 2024-06-06
ctx: Show unselectable items for unreachable Spaces

adamwg:

Avoid the surprise of seeing different Spaces in the console vs. `up ctx` by listing unreachable items for Spaces that are unreachable or for which we failed to find the ingress. Sort these items to the end of the list so that all the selectable items are at the top.

I have:

Manual testing with organizations that have many unreachable Spaces. Also tested the other menus that may have unselectable items to make sure it all looks good.

### DIFF
--- a/cmd/up/ctx/navigation.go
+++ b/cmd/up/ctx/navigation.go
@@ -16,6 +16,7 @@ package ctx
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


cmd/up/ctx/navigation.go
```diff
@@ -360,9 +379,22 @@ func (o *Organization) Breadcrumbs() string {
...
+	// Sort unselectable items last.
+	if itmI.notSelectable && !itmJ.notSelectable {
+		return false
+	}
+	if !itmI.notSelectable && itmJ.notSelectable {
+		return true
+	}
+
+	return s[i].(item).text < s[j].(item).text
+}
```
RedbackThomson, line 396: I'm not sure this is something we always want to commit to? There's a chance we'd want to put a header label, or a separator element, midway through the list and this would conflict?

Wdyt about, when we create the list of items, we have a second list of unreachable items. Then before returning the full list, we append the unreachable ones at the end?

adamwg: Yeah, I had the same thought but discovered that this worked for all the current uses of `sortedItems`. I like the idea of making the ordering explicit in in `Items()` instead - updated to do that. As these lists get more complex we might think about adding more explicit structure to them so each `Items()` method doesn't have to do the same append+sort+append dance.

RedbackThomson (approved):

Issue #575, opened 2024-07-11
invalid output "up ctx . --short" 


sttts:
### What happened?

`up ctx . --short` prints `/gke_spaces-playground_europe-west3-a_sttts-spaces-test`. But this fails:
```
$ up ctx /gke_spaces-playground_europe-west3-a_sttts-spaces-test                                                                                                                                      
up: error: ctx.Cmd.Run(): "" not found in: Upbound
```

In non-short mode, the output is `disconnected/gke_spaces-playground_europe-west3-a_sttts-spaces-test/`, which works.

Epic #

PR #577, opened 2024-07-11
ctx: unify path string

sttts:
The path string printed after every command should be working as argument, i.e. it is supposed like an absolute identifier of a context like a control plane. This was not the case before this PR. Especially the `up ctx . --short` output was its own code path with invalid string generation (e.g. lacking the `disconnected/` prefix).

Fixes: #575

### DIFF
--- a/cmd/up/ctx/navigation.go
+++ b/cmd/up/ctx/navigation.go
@@ -147,7 +147,7 @@ func (r *Root) Items(ctx context.Context, upCtx *upbound.Context, navCtx *navCon
 }
 
 func (r *Root) breadcrumbs() string {
-	return upboundRootStyle.Render("Upbound")
+	return ""
 }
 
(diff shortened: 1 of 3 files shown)


cmd/up/ctx/navigation.go
```diff
@@ -147,7 +147,7 @@ func (r *Root) Items(ctx context.Context, upCtx *upbound.Context, navCtx *navCon
 }
 
 func (r *Root) breadcrumbs() string {
-	return upboundRootStyle.Render("Upbound")
+	return ""
```
haarchri, line 150: So we need this ?

haarchri (approved):

cmd/up/ctx/navigation.go
sttts: Good call. Removed.

PR #589, opened 2024-07-25
Print the connect agent version in `up version`

jastang:

Give users a bit more info when attaching a space to Upbound.

The agent version is hardcoded for now but we may support more functions like update in the future.

I have:

```
➜  darwin_arm64 ./up version
Client:
  Version:     v0.29.0-rc.0.168.gcf62706.dirty
  Go Version:  go1.22.2
  Git Commit:  cf62706
  OS/Arch:     darwin/arm64
Server:
  Crossplane Version:        v1.15.3-up.1
  Spaces Controller Version: Upbound Cloud Managed
Connect Agent:
  Connect Agent Version: 0.0.0-408.g343d295
  ```

### DIFF
--- a/cmd/up/version/version.go
+++ b/cmd/up/version/version.go
@@ -52,8 +52,16 @@ Client:
 Server:
   Crossplane Version:	{{.CrossplaneVersion}}
   Spaces Controller Version:	{{.SpacesControllerVersion}}
-{{- end}}{{- end}}`
+{{- end}}{{- end}}
+
+{{- if ne .ConnectAgent nil}}
+Connect Agent:
+  Connect Agent Version: {{.ConnectAgent}}
+{{- end}}`
 )
+
+// TODO(jastang): remove hardcoded version if/when we support agent update.
+type connectAgentVersion string
 
 type clientVersion struct {
 	Arch      string `json:"arch,omitempty" yaml:"arch,omitempty"`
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: cmd/up/version/version.go [hunk 1]
```diff
@@ -156,6 +165,8 @@ func (c *Cmd) BuildVersionInfo(ctx context.Context, kongCtx *kong.Context, upCtx
 		v.Server.SpacesControllerVersion = versionUnknown
 	}
 
+	v.ConnectAgent = connectAgentVersion(version.AgentVersion())
```
adamwg, line 168: `version.AgentVersion` is the version this `up` CLI would install when connecting a space, which isn't necessarily what's in the connected cluster.

We could get the currently-installed version using our existing `helm` library by doing something like this (copied mostly from `up/space/connect.go`):

```go
registryURL, err := url.Parse("us-west1-docker.pkg.dev/orchestration-build/connect")
if err != nil {
	return err
}
mgr, err := helm.NewManager(kubeconfig,
	"agent",
	registryURL,
	helm.WithNamespace(agentNs),
	helm.IsOCI(),
	helm.Wait(),
	helm.Force(true),
	helm.RollbackOnError(true),
)
if err != nil {
	return err
}

agentVersion, err := mgr.GetCurrentVersion()
if err != nil {
	return err
}
```

adamwg (commented):
Not sure how valuable this is unless we actually query the agent version running in the cluster.

cmd/up/version/version.go
```diff
@@ -52,9 +54,19 @@ Client:
 Server:
   Crossplane Version:	{{.CrossplaneVersion}}
   Spaces Controller Version:	{{.SpacesControllerVersion}}
+{{- end}}{{- end}}
+
+{{- if ne .ConnectAgent nil}}{{with .ConnectAgent}}
+Connect Agent:
+  Installed Connect Agent Version: {{.InstalledVersion}}
```
phisco, line 61: Looks like you aren't printing the `desiredVersion`, is that intended?

PR #595, opened 2024-08-02
Show healthy column and read message from status for CTP list/get commands

turkenh:

With Spaces v1.6.0 we did the following changes in the control plane resource affecting the get and list command outputs:
- We no longer show `SYNCED` column, and show a new `HEALTHY` column instead.
- We store the `MESSAGE` in `status.message` rather than `internal.spaces.upbound.io/message` annotation.

This PR adapts the up cli accordingly.

This was found as a bug in the last LBOP session.

I have:

```
❯ kubectl get ctp
NAME    CROSSPLANE    READY   HEALTHY   MESSAGE                                                              AGE
test    1.15.3-up.1   True    False     Provisioning failed: cannot setup control plane API: cannot pro...   29m
test2   1.15.3-up.1   True    True

❯ go run ./cmd/up ctp list
GROUP     NAME    CROSSPLANE    READY   HEALTHY   MESSAGE                                                              AGE
default   test    1.15.3-up.1   True    False     Provisioning failed: cannot setup control plane API: cannot pro...   34m
default   test2   1.15.3-up.1   True    True                                                                           6m47s
```

### DIFF
--- a/internal/resources/controlplane.go
+++ b/internal/resources/controlplane.go
@@ -98,11 +98,15 @@ func (c *ControlPlane) GetCrossplaneVersion() string {
 }
 
 func (c *ControlPlane) GetMessage() string {
-	var ann map[string]string
-	if err := fieldpath.Pave(c.Object).GetValueInto("metadata.annotations", &ann); err != nil {
+	msg, err := fieldpath.Pave(c.Object).GetString("status.message")
+	if err != nil {
 		return ""
 	}
-	return ann["internal.spaces.upbound.io/message"]
+	return msg
+}
+
+func (c *ControlPlane) SetMessage(msg string) {
+	_ = fieldpath.Pave(c.Object).SetString("status.message", msg)
 }
 
 func (c *ControlPlane) GetAge() *time.Duration {
--- a/cmd/up/controlplane/controlplane.go
+++ b/cmd/up/controlplane/controlplane.go
@@ -38,7 +38,7 @@ import (
 )
 
 var (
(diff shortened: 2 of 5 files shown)


internal/resources/controlplane.go
```diff
@@ -98,11 +98,15 @@ func (c *ControlPlane) GetCrossplaneVersion() string {
 }
 
 func (c *ControlPlane) GetMessage() string {
-	var ann map[string]string
-	if err := fieldpath.Pave(c.Object).GetValueInto("metadata.annotations", &ann); err != nil {
+	msg, err := fieldpath.Pave(c.Object).GetString("metadata.annotations")
```
adamwg, line 101: Should this be `status.message`?

internal/resources/controlplane.go
```diff
@@ -98,11 +98,15 @@ func (c *ControlPlane) GetCrossplaneVersion() string {
...
+	if err != nil {
 		return ""
 	}
-	return ann["internal.spaces.upbound.io/message"]
+	return msg
+}
+
+func (c *ControlPlane) SetMessage(msg string) {
+	_ = fieldpath.Pave(c.Object).SetString("metadata.annotations", msg)
```
adamwg, line 109: Same here - `status.message` instead of `metadata.annotations`?

internal/resources/controlplane.go
turkenh: Good catch, fixed!

adamwg (approved):
lgtm

PR #614, opened 2024-09-04
Setup logging

sttts:
This PR sets up controller-runtime and klog logging. By default, klog verbosity is 3, but we filter by request throttling messages. With one or more of `-d` or `--debug`, or `--debug=<level>` verbosity can be increased. `-d -d -d` or `--debug=3` means klog verbosity `3+3`, to 6. With that (3), you see all the (client-go) http requests the binary is doing.

### DIFF
--- a/cmd/up/repository/repository.go
+++ b/cmd/up/repository/repository.go
@@ -49,6 +49,8 @@ func PredictRepos() complete.Predictor {
 		if err != nil {
 			return nil
 		}
+		upCtx.SetupLogging()
+
 		cfg, err := upCtx.BuildSDKConfig()
 		if err != nil {
 			return nil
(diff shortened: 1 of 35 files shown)


cmd/up/repository/repository.go
```diff
@@ -46,9 +46,13 @@ func (c *Cmd) AfterApply(kongCtx *kong.Context) error {
 func PredictRepos() complete.Predictor {
 	return complete.PredictFunc(func(a complete.Args) (prediction []string) {
 		upCtx, err := upbound.NewFromFlags(upbound.Flags{})
+		upCtx.SetupLogging()
+
 		if err != nil {
 			return nil
 		}
+		upCtx.SetupLogging()
+
```
ezgidemirel, line 55: why do we call setup twice in this function?

ezgidemirel (approved):
added one comment, other than that LGTM

cmd/up/repository/repository.go
sttts: good find. We shouldn't.

sttts: fixed

